PR #8, opened 2024-01-28
RFC: Rewrite gamelauncher in Python

R1kaB3rN:
In its current form, executing the gamelauncher script can be tedious especially when not using a modern shell that supports auto completions. For instance:

<pre>
$ WINEPREFIX=$HOME/Games/epic-games-store GAMEID=egs PROTONPATH=$HOME/.steam/steam/compatibilitytools.d/GE-Proton8-28 ./gamelauncher.sh $HOME/Games/epic-games-store/drive_c/Program Files (x86)/Epic Games/Launcher/Portal/Binaries/Win32/EpicGamesLauncher.exe -opengl -SkipBuildPatchPrereq
</pre>

By rewriting the gamelauncher script in Python, we can support optionally reading from a configuration file(s) instead which increases ease of use and provides better organization. This effectively results in this:

<pre>
[ulwgl]
prefix = "/home/foo/Games/epic-games-store"
game_id = "egs"
proton = "/home/foo/.steam/steam/compatibilitytools.d/GE-Proton8-28"
game = " /home/foo/Games/epic-games-store/drive_c/Program Files (x86)/Epic Games/Launcher/Portal/Binaries/Win32/EpicGamesLauncher.exe"
launch_opts = ["-opengl", "-SkipBuildPatchPrereq"]
</pre>

<pre>
$ gamelauncher.py --config example.toml
</pre>

Note, with this PR, we can still execute the launcher by passing environment variables. To pass game options, use` --options`
<pre>
$ WINEPREFIX=... GAMEID=... PROTONPATH=... ./gamelauncher.py --exe "/home/foo/Games/epic-games-store/drive_c/Program Files (x86)/Epic Games/Launcher/Portal/Binaries/Win32/EpicGamesLauncher.exe" --options "-opengl -SkipBuildPatchPrereq"
</pre>

Additionally, due to the rich Python ecosystem, the rewrite leaves room for future features and formal testing if needed. 

This being the case, perhaps this should be the default? I'd like to hear some feedback.

---
TODO:

- [x] Test Steam game drive functionality
- [x] Test on real games
- [x] Add support for STEAM_COMPAT_LIBRARY_PATHS
- [x] Add PROTON_VERB


R1kaB3rN:
cc @GloriousEggroll for feedback

GloriousEggroll:
I like the idea of converting it to python, but I still want to control the values by providing environment variables. They are easy to remember and don't have a hard requirement for a config file.

I would request that if you want to feed values from a config file, have the config file set envvars, and have the gamelauncher python script read those envvars like the current shell script does. Apologies if thats what its already doing, I only briefly glanced at the current code

R1kaB3rN:
> I would request that if you want to feed values from a config file, have the config file set envvars, and have the gamelauncher python script read those envvars like the current shell script does. Apologies if thats what its already doing, I only briefly glanced at the current code

Yes, you can still run the script as before. However, when passing game options users would need to wrap the value in quotes. For example:

<pre>
$ WINEPREFIX=... GAMEID=... PROTONPATH=... ./gamelauncher.py --game="/home/foo/Games/epic-games-store/drive_c/Program Files (x86)/Epic Games/Launcher/Portal/Binaries/Win32/EpicGamesLauncher.exe -opengl -SkipBuildPatchPrereq"
</pre>

 Passing a config file will be left optional.

Also note, the launcher currently depends on the [tomlkit](https://pypi.org/project/tomlkit/) library. On Arch, you can install the `python-tomlkit` for this. Though, I do not know about Fedora and other distros.

R1kaB3rN:
cc @strycore for feedback before I attempt to rewrite [ULWGL](https://github.com/Open-Wine-Components/ULWGL-launcher/blob/main/ULWGL)

strycore:
I will always approve rewriting bash stuff in Python. Just one small note: the use of TOML doesn't offer a lot of advantages versus other format but introduces an external dependency.
Using JSON would make the script run without needing installing anything first. 
For a launcher script like this one, I think it's pretty important not to introduce 3rd party components especially when they are not vital, like toml.

R1kaB3rN:
Thanks for the feedback. I substituted tomlkit with tomllib which was added in Python 3.11. This way, we depend purely on the standard library. Also, I added some tests and typing.

loathingKernel:
Nitpicking, I would prefer for `--game` to be `--exe` or something similar for clarity, as `game` can mean from a game's title or ID to the executable, while `exe` is more general (it can be some other executable in the same prefix, such as a wine tool) and highlights what the argument should be.

loathingKernel:
For commit https://github.com/Open-Wine-Components/ULWGL-launcher/pull/8/commits/b7092958dd631d80dba5d40e0b739b264c4684b3, the `ULWGL` script seems to be more or less a direct copy of `_v2-entry-point` with a few minor modifications, which comes with SteamLinuxRuntime. In that regard it might be less of a hassle to not attempt to re-write it in python, as in its current bash script form can easily be updated to any newer version for future SLR releases. I don't know anyone else's thoughts on this.

R1kaB3rN:
Reimplemented the ULWGL functionality in Python and incorporated it into the launcher script. Unit tests for the PV functionality still needs to be written so expect bugs. Manual testing (running it on real games) still needs be done too.

Feedback, review and suggestions are welcome.

GloriousEggroll:
> For commit [b709295](https://github.com/Open-Wine-Components/ULWGL-launcher/commit/b7092958dd631d80dba5d40e0b739b264c4684b3), the `ULWGL` script seems to be more or less a direct copy of `_v2-entry-point` with a few minor modifications, which comes with SteamLinuxRuntime. In that regard it might be less of a hassle to not attempt to re-write it in python, as in its current bash script form can easily be updated to any newer version for future SLR releases. I don't know anyone else's thoughts on this.

ooof hard agree here. I thought this rewrite was only pertaining to the launcher script, not ULWGL itself -- which is indeed a copy of the v2 entrypoint script valve ships. I really dont think we should be changing that, especially in case of a major update that may need porting in the future

-edit-
After re-reading the actual changes it does appear to just be a rewrite of the launcher script. Maybe there was some confusion here?

There is gamelauncher.sh -- which is our script, and there is ULWGL -- which is a 1:1 copy of valve's v2-entrypoint script that should not be changed.

If this rewrite is for gamelauncher.sh which it appears it is, I'm all for it.

R1kaB3rN:
> > For commit [b709295](https://github.com/Open-Wine-Components/ULWGL-launcher/commit/b7092958dd631d80dba5d40e0b739b264c4684b3), the `ULWGL` script seems to be more or less a direct copy of `_v2-entry-point` with a few minor modifications, which comes with SteamLinuxRuntime. In that regard it might be less of a hassle to not attempt to re-write it in python, as in its current bash script form can easily be updated to any newer version for future SLR releases. I don't know anyone else's thoughts on this.
> 
> ooof hard agree here. I thought this rewrite was only pertaining to the launcher script, not ULWGL itself -- which is indeed a copy of the v2 entrypoint script valve ships. I really dont think we should be changing that, especially in case of a major update that may need porting in the future
> 
> -edit- After re-reading the actual changes it does appear to just be a rewrite of the launcher script. Maybe there was some confusion here?
> 

Sorry, it appears that I got too carried away there and I agree -- it would be simpler/safer if we do not touch  `_v2-entry-point` (ULWGL) to cope with future changes. [And it appears that script still does change](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/commits/main/subprojects/container-runtime/common?ref_type=heads).

I think a request to rewrite it would be more appropriately requested upstream instead.

GloriousEggroll:
just to clarify the only reason we rename the v2-entrypoint script is to differentiate it from valve's because ultimately we plan to build our own version of the runtime + environment scripts. Right now it is a 1:1 copy, and even so we still need to make sure it's distinguishable from valve's version

loathingKernel:
>  After re-reading the actual changes it does appear to just be a rewrite of the launcher script. Maybe there was some confusion here?

This confusion is most likely resolved by now as I can see from the conversation. Just for clarity though, the commit I referenced before was starting to replicate and incorporate functionality found in the `ULWGL` script into `gamelauncher.py`, hence the suggestion to avoid that path and instead use the `ULWGL` in its current bash script form.

R1kaB3rN:
Applied feedback and removed the v2-entrypoint functionality as suggested. All functionality is tested except `main` and manual testing on real games is still TODO

R1kaB3rN:
Also, since this branch was based on https://github.com/Open-Wine-Components/ULWGL-launcher/commit/c3c016bcdc7d107f239e302712a1cce7c845ddc6, the new features/bug fixes from `gamelauncher.sh` are not included **yet**

R1kaB3rN:
cc @strycore for opinion on rewriting [ULWGL](https://github.com/Open-Wine-Components/ULWGL-launcher/blob/main/ULWGL), [run](https://github.com/Open-Wine-Components/ULWGL-launcher/blob/main/run) and [run-in-sniper](https://github.com/Open-Wine-Components/ULWGL-launcher/blob/main/run-in-sniper) in Python -- do you think it's worth it? 

As a result, we'd just shell out to Pressure Vessel directly rather going through two Bash scripts, thus limiting all Bash within a runtime platform.

strycore:
For ULWLGL, I thought that was the plan. For the other 2 scripts, it seems less important since those scripts are auto-generated and contain no logic so shell scripts seem perfectly suitable for this situation.

loathingKernel:
@R1kaB3rN Since I seem to always bring forth issues, let me say first that I really like the effort in general and the way it is going, I believe I have not said that yet.

That being said, it feels to me like commit d8c812559b1c20a40eee313d20308fecba9b5194 is taking a very restraining approach. I do not know what your plan is with `PROTON_VERB` yet, and since I have not tested the rewrite myself, I would suggest you consider adding the following test cases and the usefulness of their output.

```
WINEPREFIX="/<PATH_TO_FOLDER>/test/" PROTONPATH="/<HOME>l/.local/share/Steam/compatibilitytools.d/GE-Proton8-27" GAMEID="0" PROTON_VERB="runinprefix" ./gamelauncher.sh cmd.exe /c echo %LOCALAPPDATA%

> C:\users\steamuser\AppData\Local
```
This would fail due to `cmd.exe` not being a file on the Unix side of the script

```
WINEPREFIX="/<PATH_TO_FOLDER>/test/" PROTONPATH="/<HOME>/.local/share/Steam/compatibilitytools.d/GE-Proton8-27/" GAMEID="0" PROTON_VERB="ruinprefix" ./gamelauncher.sh winepath.exe -w /<PATH_TO_FOLDER>/test/drive_c//windows/notepad.exe 

> C:\windows\notepad.exe
```
This would fail due to `winepath.exe` not being a file on the Unix side of the script, and if that is fixed, it would also fail because part of the argument IS a file on the Unix side.

```
WINEPREFIX="/<PATH_TO_FOLDER>/test/" PROTONPATH="/<HOME>/.local/share/Steam/compatibilitytools.d/GE-Proton8-27/" GAMEID="0" PROTON_VERB="getnativepath" ./gamelauncher.sh 'c:\users\steamuser\'

> /<PATH_TO_FOLDER>/test/pfx/dosdevices/c:/users/steamuser/
```
This would fail because the `exe` argument is not a real file on the Unix side.

As far as I can tell from reading the code in the aforementioned commit, these will fail, either because the `exe` cannot be found (https://github.com/Open-Wine-Components/ULWGL-launcher/pull/8/commits/d8c812559b1c20a40eee313d20308fecba9b5194#diff-f56f7a9b166ff13db2e748c2f0c8693e867af851a8a8993ce8fb499f8cd284aaR148, https://github.com/Open-Wine-Components/ULWGL-launcher/pull/8/commits/d8c812559b1c20a40eee313d20308fecba9b5194#diff-f56f7a9b166ff13db2e748c2f0c8693e867af851a8a8993ce8fb499f8cd284aaR80), or the argument is indeed a file (https://github.com/Open-Wine-Components/ULWGL-launcher/pull/8/commits/d8c812559b1c20a40eee313d20308fecba9b5194#diff-f56f7a9b166ff13db2e748c2f0c8693e867af851a8a8993ce8fb499f8cd284aaR85)

I would like others to comment on the usefulness of these cases, and if it is indeed worth it to restrict them to prevent user error by removing the ability to construct such commands. For me their usefulness lies in the ability to translate Windows env variables and paths to something more meaningful for handling on the Unix side. An example of this are save file paths. Either way I am fine with either direction because I can work around this one way or another.

If I have missed or misunderstood something, by all means, feel free to correct me.

loathingKernel:
Also a bit of a code review

* I do not understand the usefulness of this syntax where you are typing the exceptions raised in the functions
```python
def _setup_pfx(path: str) -> Union[None, FileExistsError, RuntimeError]:
```
From a quick google search what I found is this [PEP 484](https://peps.python.org/pep-0484/#exceptions) which doesn't state that it is wrong but discouraged. Maybe there is a later PEP which invalidates that of which I am not aware of. I would like a bit of explanation from you on this topic, why are you doing it and what I might be missing.

* It is a personal preference but I do not like indirect returns in Python. Unlike C or C++ where it is easy to identify them, in python they are seemingly transparent to others unfamiliar with the code. I wouldn't have noticed them if the use of typing in the previous point didn't raise (pun maybe intended) my attention to how you return values from your functions.

R1kaB3rN:
> @R1kaB3rN Since I seem to always bring forth issues, let me say first that I really like the effort in general and the way it is going, I believe I have not said that yet.

Thanks, and I appreciate the feedback!

> That being said, it feels to me like commit https://github.com/Open-Wine-Components/ULWGL-launcher/commit/d8c812559b1c20a40eee313d20308fecba9b5194 is taking a very restraining approach.

This was intentional because I want to enforce users to strictly follow the project's example usage and for security -- we need to be careful as we're shelling out to a script. For instance, your first example with the `echo` command, I do not want cases like that to happen. However, I would be OK with delaying those checks for now to prioritize the merge or until more feedback.

> I do not know what your plan is with PROTON_VERB yet, and since I have not tested the rewrite myself, I would suggest you consider adding the following test cases and the usefulness of their output.

I developed on branch https://github.com/Open-Wine-Components/ULWGL-launcher/commit/c3c016bcdc7d107f239e302712a1cce7c845ddc6 so I would like to implement all of the new features/bug fixes that had been merged. Though, to prioritize this getting merge ASAP, I can address new updates in another PR. Also, I'll look into your provided test cases and try to account for them -- I'm expecting certain inputs and for them to be in a certain form. Windows paths haven't been tested and values passed to an executable's launch arguments are not accounted for. 

> From a quick google search what I found is this [PEP 484](https://peps.python.org/pep-0484/#exceptions) which doesn't state that it is wrong but discouraged. Maybe there is a later PEP which invalidates that of which I am not aware of. I would like a bit of explanation from you on this topic, why are you doing it and what I might be missing.

This decision was a personal one since functions in Python implicitly return `None` if not returning any other value. Also, the Ruff linter (default ruleset) has not alarmed me on this practice either. However, regardless of Ruff or if there exists a PEP for type hint best practices, this is would simple to add and in the future [we can enforce other practices through linting rules](https://docs.astral.sh/ruff/rules/#isort-i) and into a CI pipeline

gamelauncher_util.py
```diff
@@ -0,0 +1,73 @@
+import subprocess
+
+
+def get_steam_compat_lib():
+    """Find the library paths of a system.
+
+    Used for Steam's game drive
+    """
+    ldconfig_cmd = ["ldconfig", "-XNv"]
```
R1kaB3rN: cc @GloriousEggroll for clarification on this cmd for game drive

For this command, if I understand correctly, we want to find the user's library paths, right? If so, would it be a problem to hard code to typical paths like:
 `/usr/lib:/lib:$LD_LIBRARY_PATH`

gamelauncher_util.py
```diff
@@ -0,0 +1,73 @@
...
+def get_steam_compat_install(envar):
+    """Find the mnt point of filesystem path.
+
+    Used for Steam's game drive
+    Expects envar to be STEAM_COMPAT_INSTALL_PATH
+    """
+    findmnt_cmd = ["findmnt", "-T", envar]
+    tail_cmd = ["tail", "-n", "1"]
+    awk_cmd = ["awk", "{ print $1 }"]
+
```
R1kaB3rN: cc @GloriousEggroll 

To clarify, we want to find the mount points of a game, right? Why would Steam want to do this and which cases does this apply to?

R1kaB3rN:
> For ULWLGL, I thought that was the plan. For the other 2 scripts, it seems less important since those scripts are auto-generated and contain no logic so shell scripts seem perfectly suitable for this situation.

The expectation set here was for the gamelauncher only. Once this is merged, I plan to address rewriting ULWGL separately. 

The rewrite for it is finished, though it seems we're not all in agreement or are still unclear if it's worth it yet

gamelauncher_util.py
R1kaB3rN: cc @imLinguin for examples of expected output(s) for `$STEAM_COMPAT_LIBRARY_PATHS`

imLinguin: The way `STEAM_COMPAT_LIBRARY_PATHS` are used by Steam quite easy. Steam has the concept of libraries - locations where Steam stores prefixes, games, shaders etc.... With ULWGL we don't really have that concept and each launcher (heroic,bottles,lutris) handles that differently.

The locations in `STEAM_COMPAT_LIBRARY_PATHS` are `:` separated paths, where at least one is a parent of `STEAM_COMPAT_INSTALL_PATH`. Game drive is only used for certain games, or when explicitly enabled via env var. Proton goes through the paths in `STEAM_COMPAT_LIBRARY_PATHS` and checks if `STEAM_COMPAT_INSTALL_PATH` is included in given `LIBRARY_PATH`. If the path like that was found it's being mounted as an `S:` drive to proton's prefix. If the `STEAM_COMPAT_LIBRARY_PATHS` is not defined Proton ignores the game drive request.

That variable is only used for game drive functionality. There are two things that game drive solves.
- make path shorter (for games that can't handle paths beyond certain length)
- allow launchers (e.g Genshin Impact, The Elder Scrolls Online) to properly calculate available size (in case of separate drive or location inaccessible from the container like flatpak)

By default we added `STEAM_COMPAT_INSTALL_PATH` there however it's not reliable as the fallback for that variable is parent of .exe. As that could lead to some issues we decided to find the nearest mount point that isn't `/`. That way we allow game drive to solve point two I showcased above. The first point may still cause minor issues if path is still beyond the character limit.

sewnie:
I think it would be better to have gamelauncher to not require Python, i wouldn't want Python just to launch a game.

This is unless proton itself requires proton, if that is the case i will be fine with it.

R1kaB3rN:
> I think it would be better to have gamelauncher to not require Python, i wouldn't want Python just to launch a game.
> 
> This is unless proton itself requires proton, if that is the case i will be fine with it.

See https://github.com/ValveSoftware/Proton/blob/proton_8.0/proton

loathingKernel:
The point is not completely unreasonable, since `python` is provided by the SteamLinuxRuntime environment, which is already set up when the `proton` script is executed. The environment `gamelauncher.py` is executed in is the environment of the host, so the host has to provide python at that point. IMO it is safe to assume that python exists in the host system though or have it as a requirement.

imLinguin:
A lot of people ran proton without the container already, no issues were reported. Also python ships by default on all distros nowadays. I'd say we are safe.

R1kaB3rN:
> Also a bit of a code review
> I do not understand the usefulness of this syntax where you are typing the exceptions raised in the functions
> ```python
> def _setup_pfx(path: str) -> Union[None, FileExistsError, RuntimeError]:
> ```

Sorry, it appears that I didn't carefully read your point here. Hmm, I think having metadata like this at a glance for  functions can be really helpful. It avoids having to read the implementation and clarifies what the developer expects to happen for it, even if its at the cost of verbosity/file length. 

However, again, the Ruff linter doesn't yell at me for this practice even if I omit it. Since no static checking seems to be involved there, I'm OK with removing them unless others think its helpful. Though, sorry, [I will not explicitly return None as its redundant](https://docs.astral.sh/ruff/rules/unnecessary-return-none/).

loathingKernel:
> It avoids having to read the implementation and clarifies what the developer expects to happen for it, even if its at the cost of verbosity/file length.

I do not mind the file length nor the verbosity, personally I like good typing. TBH though I found it confusing as I was expecting return types there, not Exception types. Since linters are happy with it, and you don't use that space for anything else, such as actual return types (although having a `Union` of real and documentational return types looks odd to me), I have no further issue with it.

> Though, sorry, [I will not explicitly return None as its redundant](https://docs.astral.sh/ruff/rules/unnecessary-return-none/).

I am not sure if that refers to what I called `indirect returns`. If it does then I agree with you, explicitly returning None is bad unless it holds some merit to it, for example if the result is `Optional`.

What I referred to as `indirect return` is the passing of a dictionary as argument to the function and populating it in the function without ever returning it.  Ultimately yes, you are just passing around references to the same object, but to me returning the dictionary looks cleaner in terms of programmer intention.

R1kaB3rN:
Can confirm that running non-Steam games work, both in the TOML and environment variable usage.

Non-Steam game:

screenshot

R1kaB3rN:
Strangely though, I needed Steam installed to get this to work. I have to double check on this as it just could have been a mistake on my part, but is that an implicit requirement?

Edit:

After looking into it, it seems you need to satisfy Steam's dependencies. On Arch, it's effectively the dependencies of the steam package. If one wanted to package this, I guess they can simply use Arch's. However, it would be nice if we can precisely determine what's strictly required...

gamelauncher_util.py
R1kaB3rN: @imLinguin thanks for clarifying this for me and I just examined Proton's code to understand how it all works. And if I'm not mistaken, game drive will always be enabled for our launcher if the game resides in an external mount point that's not the root.

I was able to implement the game drive functionality in pure Python. However, I think we're still missing something here. For some reason, for one of the non-steam games I tested, it will not work when its on my external drive and only does if its in my root filesystem. I can confirm this on both for the rewrite and the original. However, it's only for that specific game, so it could that something is missing for it that we don't account for.

Also, we do not account for subvolumes which can also be mount points. Would this be a problem? If this matters, we'd need to add a dependency to handle those cases (e.g. btrfs)

R1kaB3rN: Error message:

error

shymian:
Just wanted to add an additional perspective, the inclusion of a --config game.toml launch option fits my use case _exactly_. It will work perfectly with pegasus-frontend. I hope you keep that feature.

GloriousEggroll:
How's the current state of this in terms of feature parity? I'm wondering if we should go ahead and get this merged then work on whats missing. We could always merge it alongside then replace when it's done

R1kaB3rN:
> Just wanted to add an additional perspective, the inclusion of a --config game.toml launch option fits my use case _exactly_. It will work perfectly with pegasus-frontend. I hope you keep that feature.

I'm glad to hear that. And yes, rest assured -- support for reading config files will be a core feature

R1kaB3rN:
> How's the current state of this in terms of feature parity? I'm wondering if we should go ahead and get this merged then work on whats missing. We could always merge it alongside then replace when it's done

In terms of feature parity, I'd say it's 99% complete -- I just haven't supported setting ULWGL_ID. This would be trivial to implement, but I don't quite understand the use case for it -- is it for clients like GOG? 

Other than that, I would say it's safe to merge, and it would be great to have it replace the original script, ideally, once Proton 9 releases. If you haven't already, please try it out

gamelauncher_plugins.py
```diff
@@ -0,0 +1,39 @@
...
+        paths.add(Path(os.environ["LD_LIBRARY_PATH"]).as_posix())
+
+    if env["STEAM_COMPAT_INSTALL_PATH"]:
+        paths.add(env["STEAM_COMPAT_INSTALL_PATH"])
+
+    # Hard code for now because these paths seem to be pretty standard
+    # This way we avoid shelling to ldconfig
+    paths.add("/usr/lib")
+    paths.add("/usr/lib32")
+    env["STEAM_RUNTIME_LIBRARY_PATH"] = ":".join(list(paths))
```
R1kaB3rN, line 37: cc @imLinguin for review on the game drive functionality.

We check for paths  `STEAM_COMPAT_INSTALL_PATH` up to the root, rather than in reverse. Would this a problem -- should we prefer checking a path closer to the root? 

For instance consider this case:

> /mnt/games/foo/foo.exe

Both `mnt` and `foo` are mountable. With the current implementation, `foo` (a BTRFS subvolume) will be set to `STEAM_COMPAT_INSTALL_PATH`? Would `/mnt` be preferred instead -- this is what `findmnt` would return

GloriousEggroll:
> ULWGL_ID

GAMEID = ULWGL_ID:

https://github.com/Open-Wine-Components/ULWGL-launcher/blob/a702bab465d9d9cbb6aaad711aae7c7370c4adac/ulwgl-run#L40

Proton and protontricks both check for ULWGL_ID:

https://github.com/Open-Wine-Components/ULWGL-Proton/blob/e385c2b128a0a190ebd8737b80000a36cd334c68/proton#L1695

https://github.com/Open-Wine-Components/ULWGL-protonfixes/blob/56d63fa4446f6c8befa722b3b0771cf5a0fa950a/fix.py#L20

 If anything we should drop using GAMEID instead and just expect ULWGL_ID to be passed, except GAMEID is easier to remember and type. I also feel it's a bit more user friendly.

sewnie:
`gamelauncher.sh` is missing. Update the README please.

PR #17, opened 2024-02-07
Separate self-update and download/extraction into a wrapper script for CLI use

loathingKernel:
Example implementation of https://github.com/Open-Wine-Components/ULWGL-launcher/issues/16

Create a new script `ulwgl-run-cli` which can be installed as `bin/ulgwl-run` (example in flatpak manifest) and it takes care of downloading/extracting ULWGL-launcher and ULWGL-Proton in to their respective paths. The downloads are stored temporarily in the `$HOME/.cache/ULWGL` directory (this could be `/tmp` too since they are later deleted) to avoid cluttering the current directory. The actual implementation is a direct copy of the existing one in `ulwgl-run`

### DIFF
--- a/org.openwinecomponents.ulwgl.launcher.yml
+++ b/org.openwinecomponents.ulwgl.launcher.yml
@@ -143,11 +143,11 @@ modules:
   - name: ulwgl-run
     buildsystem: simple
     build-commands:
-      - install -D ulwgl-run /app/bin/ulwgl-run
+      - install -D ulwgl-run-cli /app/bin/ulwgl-run
       - install -D ULWGL-launcher.tar.gz /app/share/ULWGL/ULWGL-launcher.tar.gz
     sources:
       - type: file
-        path: ulwgl-run
+        path: ulwgl-run-cli
       - type: file
         url: https://github.com/Open-Wine-Components/ULWGL-launcher/releases/download/0.1-RC3/ULWGL-launcher.tar.gz
         sha256: e25c4dd0636d04e7c8c534cf3c5bbdca5ae0d49f146ee8395306174700899952
--- a/toolmanifest.vdf
+++ b/toolmanifest.vdf
@@ -1,7 +1,7 @@
 // Generated file, do not edit
 "manifest"
 {
-	"commandline" "/_v2-entry-point --verb=%verb% --"
+	"commandline" "/ULWGL --verb=%verb% --"
 	"version" "2"
 	"use_tool_subprocess_reaper" "1"
 	"unlisted" "1"
--- a/ulwgl-run
+++ b/ulwgl-run
@@ -3,41 +3,15 @@
 # use for debug only.
 # set -x
 
-if [ -z "$1" ] || [ -z "$WINEPREFIX" ] || [ -z "$GAMEID" ]; then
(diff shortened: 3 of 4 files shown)


R1kaB3rN:
I'm personally against the self-update functionality as a whole. But if we do decide to keep it, I would like it to be packaged in similar way as GE-Proton -- with a version number and for the contents to be in a root directory. For instance: ULWGL-31.tar.gz with a directory structure as:

>  \\-- ULWGL-31       
         +-- pressure-vessel
         +-- ULWGL
         +-- sniper_platform_0.20231211.70175

That way, users can optionally fall back to earlier versions and we can avoid any risk when deleting files within a user's home directory

GloriousEggroll:
> I'm personally against the self-update functionality as a whole. But if we do decide to keep it, I would like it to be packaged in similar way as GE-Proton -- with a version number and for the contents to be in a root directory. For instance: ULWGL-31.tar.gz with a directory structure as:
> 
> > \-- ULWGL-31
> > +-- pressure-vessel
> > +-- ULWGL
> > +-- sniper_platform_0.20231211.70175
> 
> That way, users can optionally fall back to earlier versions and we can avoid any risk when deleting files within a user's home directory

technically the autoupdate only works for non-flatpak anyway, because we ship the ULWGL tarball into /app/share/ and it just extracts it from there to ~/.local/share/ULWGL

That being said we could take the same approach with non-flatpak -- just place the tarball in /usr/share/ULWGL and extract it to the user's ~/.local/share/ULWGL. Neither method requires auto-updating from online.

loathingKernel:
@R1kaB3rN As far as I understand purely from observation, the SteamLinuxRuntime always moves forward and it guarantees compatibility with anything built on that runtime major version. Anything build for `Sniper` for example, even if it was the first version of, will keep on running on any future version of `Sniper`. Outside of testing purposes, I do not know if keeping multiple versions of the runtime around would be required, at least until the next major version.

For example, the [toolmanifest.vdf](https://github.com/Open-Wine-Components/ULWGL-launcher/blob/main/toolmanifest.vdf) that describes the runtime only mentions of version 2

R1kaB3rN:
I think the runtimes are pretty large as well -- 600 MB for sniper. Since its pretty stable, extracting new versions of each release (ULWGL-30, ULWGL-31, etc.) would just bloat the user's directory unnecessarily. 

And Just to confirm, can we not just symlink to some of the files instead? For example, just keep the runtime, pressure vessel and other compat tool directories in `/usr/share/ULWGL `while extracting only the launcher scripts in `~/.local/share/ULWGL`? That way, we limit writing to the user's home directory for each new release.

And after some quick testing, I can confirm this works. Also, since TOML files are supported in the upcoming launcher rewrite, we have a way to keep track of state.

GloriousEggroll:
@R1kaB3rN are you on Discord? Need to contact you but your github profile is private

GloriousEggroll:
going to merge this for now, i think our focus should shift on getting the python launcher in shape

R1kaB3rN:
> @R1kaB3rN are you on Discord? Need to contact you but your github profile is private

@GloriousEggroll Yes, I'm already on Discord and have mentioned you

PR #21, opened 2024-02-09
Support symlink of current unix username to steamuser

### LINKED ISSUE #15: Symlinking current unix username to `steamuser`
As a result of Valve's wine hardcoding the wine user name to `steamuser`, prefixes created with other wine versions, even `wine-ge` will lose access to the wine user directory named after the current unix user. It will also cause games to lose access to user-local files in case the user choose another runner not based on proton/ulwgl.

Since the general consensus is to keep compatibility with wine, for minimum disruption, `ulwgl-run` should also create a symlink named  `drive_c/users/<unix_username>` pointing to `drive_c/users/steamuser`.

R1kaB3rN:
Closes https://github.com/Open-Wine-Components/ULWGL-launcher/issues/15

Supports the creation of a symbolic link of the current user to the steamuser directory or vice versa a la Lutris. In all other cases, we simply do nothing and continue. 

cc @loathingKernel

### DIFF
--- a/ulwgl_run.py
+++ b/ulwgl_run.py
@@ -13,6 +13,7 @@ from ulwgl_dl_util import get_ulwgl_proton
 from ulwgl_consts import Level
 from ulwgl_util import msg
 from ulwgl_log import log, console_handler, debug_formatter
+from ulwgl_util import UnixUser
 
 verbs: Set[str] = {
     "waitforexitandrun",
@@ -86,6 +87,11 @@ def set_log() -> None:
 def setup_pfx(path: str) -> None:
     """Create a symlink to the WINE prefix and tracked_files file."""
     pfx: Path = Path(path).joinpath("pfx").expanduser()
+    steam: Path = Path(path).expanduser().joinpath("drive_c/users/steamuser")
+    user: UnixUser = UnixUser()
+    wineuser: Path = (
+        Path(path).expanduser().joinpath(f"drive_c/users/{user.get_user()}")
+    )
 
     if pfx.is_symlink():
         pfx.unlink()
@@ -94,6 +100,40 @@ def setup_pfx(path: str) -> None:
         pfx.symlink_to(Path(path).expanduser())
 
     Path(path).joinpath("tracked_files").expanduser().touch()
+
+    # Create a symlink of the current user to the steamuser dir or vice versa
+    # Default for a new prefix is: unixuser -> steamuser
+    if (
+        not wineuser.is_dir()
+        and not steam.is_dir()
+        and not (wineuser.is_symlink() or steam.is_symlink())
+    ):
+        # For new prefixes with our Proton: user -> steamuser
+        steam.mkdir(parents=True)
+        wineuser.unlink(missing_ok=True)
+        wineuser.symlink_to("steamuser")
+    elif wineuser.is_dir() and not steam.is_dir() and not steam.is_symlink():
+        # When there's a user dir: steamuser -> user
+        # Be sure it's relative
+        steam.unlink(missing_ok=True)
+        steam.symlink_to(user.get_user())
+    elif not wineuser.exists() and not wineuser.is_symlink() and steam.is_dir():
+        wineuser.unlink(missing_ok=True)
+        wineuser.symlink_to("steamuser")
+    else:
+        paths: List[str] = [steam.as_posix(), wineuser.as_posix()]
+        log.warning(
+            msg(
+                f"Skipping link creation for prefix: {pfx}",
+                Level.WARNING,
+            )
+        )
+        log.warning(
+            msg(
+                f"Following paths already exist: {paths}",
+                Level.WARNING,
+            )
+        )
 
 
 def check_env(
--- a/ulwgl_test.py
+++ b/ulwgl_test.py
@@ -873,7 +873,7 @@ class TestGameLauncher(unittest.TestCase):
             rmtree(new_link.as_posix())
 
     def test_setup_pfx_symlinks(self):
-        """Test _setup_pfx for valid symlinks.
+        """Test setup_pfx for valid symlinks.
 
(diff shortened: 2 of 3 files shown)


gamelauncher.py
```diff
@@ -49,13 +49,36 @@ def parse_args() -> Namespace:  # noqa: D103
...
+
     if not (Path(path + "/pfx")).expanduser().is_symlink():
         # When creating the symlink, we want it to be in expanded form when passed unexpanded paths
-        # Example: pfx -> /home/.wine
+        # Good: pfx -> /home/foo/.wine
+        # Bad: pfx -> ~/.wine
         Path(path + "/pfx").expanduser().symlink_to(Path(path).expanduser())
     Path(path + "/tracked_files").expanduser().touch()
 
+    # Create a symlink of the current user to the steamuser dir for Steam functionality
```
loathingKernel: There is a chance, because we don't know what the user has done previously, that they will both exist as directories. I think it would be fine to just fall through without doing anything if both `steamuser` and `<unix_username>` folders exist, and just present the user with a warning of some sort.

Even if we don't want to explicitly check because it can get messy, we can just catch the error, ignore it and continue like it never happened at long as `steamuser` exists already.

Edit: I don't know if I am being overly cautious here.

gamelauncher.py
```diff
@@ -49,13 +49,36 @@ def parse_args() -> Namespace:  # noqa: D103
 
 
 def _setup_pfx(path: str) -> None:
-    """Create a symlink to the WINE prefix and tracked_files file."""
+    """Create a symlink to the WINE prefix and tracked_files file.
+
+    Also, create a symlink of steamuser to the Unix username
+    """
+    username: str = Path().home().name
```
loathingKernel: Could `os.environ["USER"]` be used here instead? It would normally be the same though but there is no guarantee.

R1kaB3rN: It could. But between the two, I have more confidence that using `pathlib` is a stronger guarantee

gamelauncher.py
R1kaB3rN: Thanks for catching that case, and it should fairly simple to handle it. Also, no worries. It's important to get this right, hence this being a draft, since we're making a change to the user's home directory which may contain their game save data.

Personally though, I'm sorta on the fence of this feature because I think we're being too intrusive and it can sorta encourage the practice of re-using WINE prefixes -- which I'm personally against. In addition to overwriting some of the WINE prefix files with our own, we're deciding for the user's on how their home directory should be structured. For me though, I wouldn't mind as long as my save data is safe.

Alternatively, if we know ahead of time it's not a ULWGL created prefix, then we could try to reflink it to a new one.

R1kaB3rN: Besides Steam cloud saves, what else do we gain from this? If implementing this by default, I think this needs to be warranted with enough benefits as we could just simply do nothing and warn the user or hide this change behind a CLI option

gamelauncher.py
loathingKernel: The lack of guarantee is that the home directory might not match the user name. You can have your HOME directory with a different stem than your username. It's is unlikely generally but there is nothing blocking it from happening, see `useradd --home-dir`.

As far as I can tell, because Wine source is not easy to navigate, the username is assigned [here](https://github.com/wine-mirror/wine/blob/master/dlls/ntdll/unix/loader.c#L384). In my opinion, inferring it from `Path.home` is the least accurate one. Replicating Wine's way of getting the user id, we can either base it on the `USER` env variable, or the password database through the [`pwd` module](https://docs.python.org/3/library/pwd.html), for example `pwd.getpwuid(os.getuid())[0]`.

Or is a more simple way, trust python and get it with [`os.getlogin()`](https://docs.python.org/3/library/os.html#os.getlogin) or even better [`getpass.getuser()`](https://docs.python.org/3/library/getpass.html#getpass.getuser).

As a side-note, recently I had a discussion on Reddit where the poster was overriding the `USER` env variable to enforce the username for a game, to work around an issue with the game. There is no end to what silly and unexpected things users might try to do to get games to work, and most of the time they are for the better, as it helps strengthen the software. I am not saying that it is something ULWGL should support, quite the opposite actually, but it might be worth considering if it should be explicitly disallowed for consistency's sake. In that regard, using `USER` is also just as problematic as getting it from `Path.home`. Maybe it should solely depend on something that won't change, i.e. `pwd` module?

gamelauncher.py
loathingKernel: > Personally though, I'm sorta on the fence of this feature because I think we're being too intrusive and it can sorta encourage the practice of re-using WINE prefixes

I agree with you on this one, but according to a previous discussion on Discord around symlinking `<path>/<WINEPREFIX>/pfx` to `<path>/<WINEPREFIX>`, the general consensus was that users will be users and they will mix and match prefixes to wine versions anyway, and keeping compatibility is more important.

> Besides Steam cloud saves, what else do we gain from this?

We do not gain steam cloud saves in ULWGL with this change. Steam cloud saves is part of the reason Valve opted for hardcoding the user folder to `steamuser`. This is also the reason we cannot patch out this behavior from Proton builds as they are meant to be also usable through Steam AFAIK. We also cannot patch this behavior into other versions of Wine, because they are not controlled by us necessarily, for example Wine from distribution repositories. Hence the need to handle it on the filesystem level.

This discussion stems directly from there, as in this case we need to have common saves files and game configuration in both ULWGL based runners and Wine bases runners, since the prefix is shared.

>  we're deciding for the user's on how their home directory should be structured.

Only their wine home directory, which as long as it works, it shouldn't be something they will ever notice. And as long as we do not overwrite anything, there shouldn't be a problem either.

The one problem I can think of is that a few years back Wine changed its wine home folder structure from Windows XP to Windows Vista, I think this was from Wine 6 to Wine 7, which is ample time to not be considered an issue any more.

Issue #4, opened 2024-01-24
TODO: runtime container customization


GloriousEggroll:
- use our own runtime container based on valve's instead of using valve's, this way we can add/remove things from the container as needed

Issue #1, opened 2024-01-22
TODO: build environment configuration


GloriousEggroll:
- build binaries from source instead of copying from Valve's SteamLinuxRuntime_sniper (original source is at https://gitlab.steamos.cloud/steamrt/steam-runtime-tools)

orowith2os:
I'd also like to add pre/suffixes for game IDs; if a game doesn't have any pre/sufffix, it can be assumed that using the plain protonfix is fine. Otherwise, if it has a pre/suffix attached, it needs a more specific fix in order to work. This may be useful for platforms on which distributors provide different variants of their games, or different versions. 

Which, speaking of, versioning protonfixes seems like a good idea as well. This would be worthy of more discussion.

crestcere:
It would be great if we could use it with flatpak. I don't want to deal with 32bit libraries.

imLinguin:
@crestcere pressure-vessel is based on bwrap, this essentially provides same behaviour as flatpak

cybik:
Silly idea, but beyond [pre/suf]fix, could there be a "multiple options" feature?

Example: let's say ULWGL-6384 aka Genshin Impact (Launcher), is in the database. A player could want to use the Google Play PC variant instead of the Standard Launcher for reasons all their own. In such a case, the ULWGL "profile" could fetch the appropriate installer, or fetch the generic installer and apply specific config.ini changes to make the local install into a Google Play PC-enabled install.

There could be multiple installer variants for other titles as well, I just have this example on hand.

bohanubis:
can wayland driver be enabled by an enviroment variable if this lands

GloriousEggroll:
> can wayland driver be enabled by an enviroment variable if this lands

please stop spam asking about this. we already know it exists, and I already know what is needed. We will add it when Valve adds it. There are good reasons why this is not being rushed on their side.

PR #22, opened 2024-02-10
RFC: Rewrite _v2-entry-point in Python

R1kaB3rN:
Currently, our launchers shells out to two shell scripts -- _v2-entry-point ([ULWGL](https://github.com/Open-Wine-Components/ULWGL-launcher/blob/main/ULWGL)), [run](https://github.com/Open-Wine-Components/ULWGL-launcher/blob/main/run) or [run-in-sniper.](https://github.com/Open-Wine-Components/ULWGL-launcher/blob/main/run-in-sniper) This doesn't seem to be a problem and is perfectly fine to just leave it, however this PR demonstrates that it doesn't have to be that way as it's fairly trivial to implement in Python. Whether the benefits, if there is any, warrants the rewrite is left open to debate. Until there's a consensus, this PR will be left in a draft state.

Related to https://github.com/Open-Wine-Components/ULWGL-launcher/pull/8

### DIFF
--- a/ulwgl.py
+++ b/ulwgl.py
@@ -7,10 +7,9 @@ import sys
 from pathlib import Path
 import tomllib
 from typing import Dict, Any, List, Set
-import gamelauncher_plugins
+import ulwgl_plugins
 from re import match
-
-# TODO: Only set the environment variables that are not empty
+from io import TextIOWrapper
 import subprocess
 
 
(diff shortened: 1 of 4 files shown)


loathingKernel:
This might complicate #1 and #4, since these scripts are part of the SteamLinuxRuntime distribution, and they will have to be replaced after it is built. I do not think it is worth the maintance effort. and the potential incompatibilities.

R1kaB3rN:
> This might complicate https://github.com/Open-Wine-Components/ULWGL-launcher/issues/1 and https://github.com/Open-Wine-Components/ULWGL-launcher/issues/4, since these scripts are part of the SteamLinuxRuntime distribution, and they will have to be replaced after it is built. I do not think it is worth the maintance effort. and the potential incompatibilities.

With this, those scripts will just be ignored so all the client has to do is execute our launcher. I agree though, it's not clear to me if a custom solution will be worth the maintenance effort...

loathingKernel:
> With this, those scripts will just be ignored so all the client has to do is execute our launcher.

Oh right, got it, I didn't think that through.

PR #23, opened 2024-02-11
Change usage for Python rewrite

R1kaB3rN:
This changes the CLI usage of the gamelauncher from:
`gamelauncher.py --exe foo.exe --options '-opengl' `-> `gamelauncher.py foo.exe -opengl`

Moreover, the following options will be deprecated/removed: `--exe`, `--verb`, `--store`. To use them, you must set their corresponding environment variables (`PROTON_VERB`, `STORE`)

Reading from configuration files will continue to be supported.


GloriousEggroll:
merging -- this is working and has feature parity with the posix script -- thanks again @R1kaB3rN!

PR #26, opened 2024-02-13
Add Github Action for ulwgl_*.py files

### LINKED ISSUE #10: TODO: include Github Actions
Include Github Actions for the project to lint and run tests on new changes for the gamelauncher.

Depends on the new rewrite https://github.com/Open-Wine-Components/ULWGL-launcher/pull/8 being merged.

R1kaB3rN:
Closes https://github.com/Open-Wine-Components/ULWGL-launcher/issues/10


.github/workflows/ulwgl-python.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Install dependencies
+      run: |
+        python3 -m pip install --upgrade pip
+    - name: Lint ulwgl_*.py files with Ruff
+      run: |
+        pip install ruff
+        ruff --output-format github ulwgl_plugins.py ulwgl_run.py ulwgl_test.py
+    - name: Format ulwgl_*.py files with Ruff
+      run: |
+        ruff format ulwgl_plugins.py ulwgl_run.py ulwgl_test.py
```
loathingKernel: What is the intention with this? Formatting won't be applied to the repo files without committing the files back to the repo, is this purely for visual inspection?

Could this be elevated to an error to block the PR, if the file is not properly formatted? For example, if formatting results in modified content? Can be checked with git and some scripting.

loathingKernel:
Consider using a matrix for python versions like the following
```yml
    strategy:
      matrix:
        version: ['3.11', '3.12']
```
And later reference the version in this way
```yml
        uses: actions/setup-python@v5
        with:
          python-version: ${{ matrix.version }}
```
In the future more python versions will need to be checked and this makes it easy to add. You could also add previous versions in case some old Ubuntu version forces us making `tomllib` optional.

.github/workflows/ulwgl-python.yml
R1kaB3rN: Oh, I wasn't aware of that. My intention is to apply consistent formatting for these files, and I observed  `ruff format` can return an exit status  in the case it's used before the Linter. I haven't observed any errors that it can uniquely find yet, so I can remove this -- I really only care about the Linter

But I would still like consistent formatting to be applied, if possible. Is there a way to do that or is it only pre-commit to Github?

PR #28, opened 2024-02-14
ulwgl_run: recreate the symlink each run

### LINKED ISSUE #27: setup_pfx in ulwgl_run.py a bit too assumptious
https://github.com/Open-Wine-Components/ULWGL-launcher/blame/4f64389ec4167b3d2ab4bc381fe4c1f4c29e440f/ulwgl_run.py#L45-L52

Currently, the above lines work quite different from the shell version, because so far the WINEPREFIX would be pointing to the proton root with a pfx folder in it. Running the new Python version now gives the following error (note that $WINEPREFIX is the placeholder I put here for simplicity):

```
Traceback (most recent call last):
  File "/home/ben/.local/share/ULWGL/ulwgl-run", line 319, in <module>
    main()
  File "/home/ben/.local/share/ULWGL/ulwgl-run", line 303, in main
    setup_pfx(env["WINEPREFIX"])
  File "/home/ben/.local/share/ULWGL/ulwgl-run", line 51, in setup_pfx
    Path(path + "/pfx").expanduser().symlink_to(Path(path).expanduser())
  File "/usr/lib/python3.11/pathlib.py", line 1198, in symlink_to
    os.symlink(target, self, target_is_directory)
FileExistsError: [Errno 17] File exists: '$WINEPREFIX' -> '$WINEPREFIX/pfx'
```

However, I can live with this incompatibility, since this is still prerelease. What buggs me about it is the symlink and especially the absolute path in the created symlink. In addition, it seems to not recreate the symlink when it exists, which will result in a bad experience and errors when users move or copy their WINEPREFIXes. I would love to keep the simplicity of just moving a wine prefix folder somewhere and just adjusting the WINEPREFIX environment variable without manually recreating the correct symlink for `pfx`.

R1kaB3rN:
Closes https://github.com/Open-Wine-Components/ULWGL-launcher/issues/27

### DIFF
--- a/ulwgl_run.py
+++ b/ulwgl_run.py
@@ -44,11 +44,9 @@ example usage:
 
 def setup_pfx(path: str) -> None:
     """Create a symlink to the WINE prefix and tracked_files file."""
-    if not (Path(path + "/pfx")).expanduser().is_symlink():
-        # When creating the symlink, we want it to be in expanded form when passed unexpanded paths
-        # Example: pfx -> /home/foo/.wine
-        # NOTE: When parsing a config file, an error can be raised if the prefix doesn't already exist
-        Path(path + "/pfx").expanduser().symlink_to(Path(path).expanduser())
+    if Path(path + "/pfx").expanduser().is_symlink():
+        Path(path + "/pfx").expanduser().unlink()
+    Path(path + "/pfx").expanduser().symlink_to(Path(path).expanduser())
     Path(path + "/tracked_files").expanduser().touch()
 
 
--- a/ulwgl_run.py
+++ b/ulwgl_run.py
@@ -44,9 +44,14 @@ example usage:
 
 def setup_pfx(path: str) -> None:
     """Create a symlink to the WINE prefix and tracked_files file."""
-    if Path(path + "/pfx").expanduser().is_symlink():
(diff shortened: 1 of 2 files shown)


bziemons:
Thanks for the quick fix. I guess the remaining question from #27 is, whether the symlink is actually necessary and a good idea.

R1kaB3rN:
No problem and thanks for reporting this.

> I guess the remaining question from https://github.com/Open-Wine-Components/ULWGL-launcher/issues/27 is, whether the symlink is actually necessary and a good idea.

Creating the symbolic link results in a WINE prefix to be structured just like WINE's, and is a good idea as it works around the problem of root directory of the prefix being `pfx` instead of the WINEPREFIX specified by the user. Also, without it, I believe a new prefix will be created each time if the path's stem isn't `pfx` which is definitely not good

bziemons:
Yeah, this always was a bit of a pain when using Proton. They use the same environment variable WINEPREFIX and kinda do something else. However, I am not sure what the reasoning is behind creating the pfx folder in the first place and I worry it would be unexpected for Proton, if there was a directory/file clash in WINEPREFIX vs. WINEPREFIX/pfx. That's why I am unsure about this.

PR #30, opened 2024-02-15
Add download functionality to ulwgl_run

### LINKED ISSUE #29: Missing auto download of proton in python rewrite
The python version doesn't automatically download proton when PROTONPATH is missing like in the old bash version, ulwgl-run-cli file still exists but is not executed automatically

R1kaB3rN:
Closes https://github.com/Open-Wine-Components/ULWGL-launcher/issues/29 and https://github.com/Open-Wine-Components/ULWGL-launcher/issues/16 

With the current rewrite, when the user does not specify the `$PROTONPATH,` no download of the latest Proton is made on their behalf and this PR aims add that same functionality.

When no Proton is specified, we only attempt to download the latest Proton if a Proton cannot be found in the Steam compat tools directory. Otherwise, referring to `$HOME/.cache/ULWGL` if something wrong happens when downloading or if the user is offline.

Concretely the order is:

> Steam compat -> Cache -> Download -> Cache

Effectively, new installations will always download the latest Proton, while existing ones will simply be warned to the console of a later version. Downloading is avoided if the latest version already exists in the cache, and the cache will be used as a last resort (e.g. download fails, digest mismatch, etc.) to set the variable. When we're unable to find an existing Proton or download one, we simply terminate the script as we there's no point using it without specifying Proton.

TODO:
- [x]  Change GE-Proton -> ULWGL-Proton
- [x]  Add tests

### DIFF
--- a/ulwgl_dl_util.py
+++ b/ulwgl_dl_util.py
@@ -111,9 +111,7 @@ def _fetch_proton(
     """Download the latest ULWGL-Proton and set it as PROTONPATH."""
     hash, hash_url = files[0]
     proton, proton_url = files[1]
-    proton_dir: str = proton[
-        : proton.find(".tar.gz")
-    ]  # Proton dir without suffixes/dashes
+    proton_dir: str = proton[: proton.find(".tar.gz")]  # Proton dir
 
     # TODO: Parallelize this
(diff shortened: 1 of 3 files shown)


GloriousEggroll:
LGTM!

PR #31, opened 2024-02-19
Do not silence subprocess's exit code and avoid polluting `stdout`

loathingKernel:
This PR aims to address the following issues.

* Capture and return the exit code from the `subprocess` to the caller in case ULWGL itself was executed successfully. The caller might use the exit code to identify issues in the execution of the subprocess or ULWGL itself. Note: the ruff check was explicitly disabled because the exception at [line](https://github.com/Open-Wine-Components/ULWGL-launcher/pull/31/commits/f0f557688c4f489f5aeb61b7c26227bafca4af0e#diff-3bad07da2f6fd387cac60a26fa0bf37268bf5a8b4b63e03a05fa30abbfc04385R338) is used for logging, similar to the patterns ruff [ignores by default]( https://docs.astral.sh/ruff/rules/blind-except/)

* ULWGL should use `stderr` only for its own logging to avoid polluting `stdout` where the caller might expect the result of the argument command. Examples of such internal commands are `winepath.exe` and `reg.exe` where the output can be used by the caller.

@R1kaB3rN Since you are actively working on the python code and they might cause conflicts, feel free to either cherry-pick or merge by hand any of these changes if you agree with them. I also updated a few long strings to not exceed the 100th column to make the code slightly easier to parse.

### DIFF
--- a/ulwgl_run.py
+++ b/ulwgl_run.py
@@ -2,6 +2,7 @@
 
 import os
 import argparse
+from traceback import print_exception
 from argparse import ArgumentParser, Namespace
 import sys
 from pathlib import Path
@@ -281,7 +282,7 @@ def build_command(
     return command
 
 
-def main() -> None:  # noqa: D103
+def main() -> int:  # noqa: D103
     env: Dict[str, str] = {
         "WINEPREFIX": "",
         "GAMEID": "",
@@ -327,8 +328,12 @@ def main() -> None:  # noqa: D103
         os.environ[key] = val
 
     build_command(env, command, opts)
-    subprocess.run(command)
+    return subprocess.run(command).returncode
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        sys.exit(main())
+    except Exception as e:  # noqa: BLE001
+        print_exception(e)
+        sys.exit(1)
--- a/ulwgl_run.py
+++ b/ulwgl_run.py
@@ -35,13 +35,13 @@ example usage:
 
     if not sys.argv[1:]:
         err: str = "Please see project README.md for more info and examples.\nhttps://github.com/Open-Wine-Components/ULWGL-launcher"
-        parser.print_help()
+        parser.print_help(sys.stderr)
         raise SystemExit(err)
 
     if sys.argv[1:][0] in opt_args:
         return parser.parse_args(sys.argv[1:])
 
-    return (sys.argv[1], sys.argv[2:])
+    return sys.argv[1], sys.argv[2:]
 
 
 def setup_pfx(path: str) -> None:


ulwgl_run.py
```diff
@@ -16,15 +17,15 @@
...
+    usage: str = (
+        "example usage:\n"
+        f"  WINEPREFIX= GAMEID= PROTONPATH= {exe} /home/foo/example.exe\n"
+        f"  WINEPREFIX= GAMEID= PROTONPATH= {exe} /home/foo/example.exe -opengl\n"
+        f"  WINEPREFIX= GAMEID= PROTONPATH= {exe} \"\"\n"
+        f"  WINEPREFIX= GAMEID= PROTONPATH= PROTON_VERB= {exe} /home/foo/example.exe\n"
+        f"  WINEPREFIX= GAMEID= PROTONPATH= STORE= {exe} /home/foo/example.exe\n"
+        f"  {exe} --config /home/foo/example.toml\n"
+    )
     parser: ArgumentParser = argparse.ArgumentParser(
```
R1kaB3rN: I'd really like to automate style changes like this via a formatter (e.g. ruff format) then have a bot just format the files on every push. IIRC, ruff format can be configured to set a line-length. That way, we can just focus on the logic and developing core functionality.

Aside from this, personally, I think having those newline characters added makes the string look worse...

loathingKernel: > then have a bot just format the files on every push. IIRC, ruff format can be configured to set a line-length.

I understand that, but the bot will fail sometimes. It might not split at the newline, which is the more convenient place to split. I do agree though that consistent formatting is key.

> Aside from this, personally, I think having those newline characters added makes the string look worse...

IMO it is a tradeoff, you are either breaking the indentation by using a docstring or you get the newline characters. Personally I find breaking indentation worse, but this is not a dealbreaker in any way. `print` also supports the [`sep` argument](https://docs.python.org/3/library/functions.html#print) which can do the same thing, for example 
```python
print("test", "test2", sep="\n")
```
will print
```
test
test2
```
which could help in some other usages in the code.

Is this something you think I should update the PR with?

R1kaB3rN: Sorry, but I believe the best way forward in the case that when strings are too long is to delegate formatting to either Ruff or possibly another formatting tool. 

We shouldn't really have to manually do this.

loathingKernel: That's fair. I will remove that commit then.

ulwgl_run.py
```diff
@@ -327,8 +334,16 @@ def main() -> None:  # noqa: D103
...
 if __name__ == "__main__":
-    main()
+    try:
+        ret = main()
+    except Exception as e:  # noqa: BLE001
+        traceback.print_tb(e.__traceback__, file=sys.stderr)
+        print(f"{e}", file=sys.stderr)
+        sys.exit(1)
+    else:
+        sys.exit(ret)
```
R1kaB3rN: Please correct me if I am wrong, but `print_tb` just formats the traceback nicely compared to simply raising the exception, correct?

loathingKernel: Yes and no, `print_tb` is the way to print the traceback from the exception object, but ideally here, if logging is implemented, the `print_tb` and `print` calls should be replaced with `logging.excpetion`. This was an intermediate solution without using `logging`

The point here is to not simply raise the exception and exit due to it, but handle it, print the information and exit with an exit code of our choice.

ulwgl_run.py
```diff
@@ -2,6 +2,7 @@
 
 import os
 import argparse
+import traceback
 from argparse import ArgumentParser, Namespace
```
R1kaB3rN, line 6: I think we should try to import what we need here for now.

> from traceback import print_tb

Or probably better, just [print_exception](https://docs.python.org/3/library/traceback.html#traceback.print_exception)

> from traceback import print_exception

ulwgl_run.py
```diff
@@ -327,8 +328,16 @@ def main() -> None:  # noqa: D103
         os.environ[key] = val
 
     build_command(env, command, opts)
-    subprocess.run(command)
+    proc = subprocess.run(command)
+    return proc.returncode
```
R1kaB3rN, line 332: Prefer:

> return subprocess.run(command).returncode

ulwgl_run.py
loathingKernel: agreed

ulwgl_run.py
```diff
@@ -327,8 +328,16 @@ def main() -> None:  # noqa: D103
...
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        ret = main()
```
R1kaB3rN: Prefer:

> sys.exit(main())

ulwgl_run.py
```diff
@@ -327,8 +328,16 @@ def main() -> None:  # noqa: D103
...
+    except Exception as e:  # noqa: BLE001
+        traceback.print_tb(e.__traceback__, file=sys.stderr)
```
R1kaB3rN: Standard error  is the default, right? Also, when using print_exception:
> print_exception(e)

ulwgl_run.py
```diff
@@ -327,8 +328,16 @@ def main() -> None:  # noqa: D103
...
+        print(f"{e}", file=sys.stderr)
```
R1kaB3rN: Because of the above line, we shouldn't have to print twice, right?

loathingKernel: In case of `print_tb` it was required to get the message. It was printing only the traceback. In case `print_exception` behaves differently, I will remove it.

ulwgl_run.py
loathingKernel: I do not really like it this way, without good reason though. Ideally I would assign `ret` and exit in a `finally` block but since I do not really have a good reason for it, outside of aesthetics I will do it the suggested way

ulwgl_run.py
R1kaB3rN: No, I believe this is unnecessary and I've confirmed this when using `print_exception` on a syntax error

ulwgl_run.py
loathingKernel: Yes, I like to be explicit sometimes with no good reason really. Removed it.

PR #33, opened 2024-02-20
Add version flag to launcher

R1kaB3rN:
Adds a version flag to launcher and prints the release version of the CLI as well as its associated tooling that are defined in the TOML config file -- ULWGL_VERSIONS.toml.

Example of ULWGL_VERSIONS.toml:

<pre>
# Required. We expect this table
[ulwgl.versions]
# Required. We expect this key
launcher = "1.0"
# All other key value pairs can be anything and we just print them all
reaper = "1.0" 
pressure_vessel = "1.0"
runtime = "sniper"
</pre>

NOTE: This file needs to be included in the release.

For the config. we check the current directory for the file and `/usr/share/ULWGL`. In case of absence, nothing will be printed.

As a result, versions for our tools are centralized and we avoid hard coding version names in scripts.  In addition, the version config files can be autogenerated via a Github Action

### DIFF
--- a/ulwgl_run.py
+++ b/ulwgl_run.py
@@ -7,7 +7,7 @@ from argparse import ArgumentParser, Namespace
 import sys
 from pathlib import Path
 import tomllib
-from typing import Dict, Any, List, Set, Union, Tuple, NoReturn
+from typing import Dict, Any, List, Set, Union, Tuple
 import ulwgl_plugins
 from re import match
 import subprocess
@@ -45,8 +45,8 @@ example usage:
     return sys.argv[1], sys.argv[2:]
 
 
-def print_versions(paths: List[Path]) -> NoReturn:
-    """Print the versions of this launcher and all of its associated tools declared in the config file.
+def get_versions(paths: List[Path]) -> str:
+    """Return the version of this launcher and all of its associated tools declared in the config file.
 
     NOTE: The following table is required: [ulwgl.versions]
     """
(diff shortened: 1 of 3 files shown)


ulwgl_run.py
```diff
@@ -312,6 +348,16 @@ def main() -> int:  # noqa: D103
     args: Union[Namespace, Tuple[str, List[str]]] = parse_args()
     opts: List[str] = None
 
+    if isinstance(args, Namespace) and getattr(args, "version", None):
+        paths: List[Path] = [
+            Path(__file__).parent.joinpath("ULWGL_VERSIONS.toml"),
+            Path("/usr/share/ULWGL/ULWGL_VERSIONS.toml"),
+        ]
+
+        print(get_versions(paths), file=sys.stderr)
```
loathingKernel, line 357: Is this just for visual inspection or something that a client could use to make choices?

If this is the latter, this might be a case where the result should be to `stdout`, since this is an atomic operation in the sense that nothing else is going to happen when the version is requested, and it might be useful to the caller.

In that case, maybe also consider outputting it in an easily parse-able format like json.

GloriousEggroll:
lgtm just leaving open in case toml handling changes due to https://github.com/Open-Wine-Components/ULWGL-launcher/issues/37

Feel free to merge once that's resolved.

ulwgl_run.py
R1kaB3rN: Good point, and I could make the change to stdout instead for it to be parsed, but that wasn't my intention. This information is intended for humans, so the use case for external application consumption is out of scope

loathingKernel: I see, then it is fine to go to `stderr`, thanks for the clarification.

PR #38, opened 2024-02-20
Make parsing configuration files a plugin

### LINKED ISSUE #37: Runner using libs not available in Python < 3.11 causing `ModuleNotFoundError`
`ulwgl_run.py` is using 
https://github.com/Open-Wine-Components/ULWGL-launcher/blob/33bd82851aba4106851523cd952511ead0057da2/ulwgl_run.py#L9

which is not part of the Python stdlib [prior to 3.11](https://docs.python.org/3/library/tomllib.html).
```
  File "/home/zocker/testing/runtime/ULWGL-launcher/ulwgl-run", line 9, in <module>
    import tomllib
ModuleNotFoundError: No module named 'tomllib'
```
Ubuntu 22.04 is shipping Python 3.10 by default.

R1kaB3rN:
Closes https://github.com/Open-Wine-Components/ULWGL-launcher/issues/37

This PR separates the config file functionality as a plugin from the core program to make it run on some distributions (e.g. Ubuntu LTS). `tomllib` will be dynamically imported, optional and in the event the user tries to use the `--config` flag, an error will be raised to warn them of the Python requirement.

cc @loathingKernel for the Github Action


loathingKernel:
That was fast! I would suggest to add all versions from 3.7 onwards since lutris supports them.

.github/workflows/ulwgl-python.yml
```diff
@@ -32,5 +33,10 @@ jobs:
         pip install ruff
         ruff --output-format github ulwgl_*.py
     - name: Test with unittest
+      if: matrix.python-version == 3.10
```
loathingKernel: I think this should be referenced as `${{ matrix.version }}` for it to work.

R1kaB3rN:
> That was fast! I would suggest to add all versions from 3.7 onwards since lutris supports them.

We could test that far, but is it fine to use Lutris as our baseline? They must have different requirements to test that far

.github/workflows/ulwgl-python.yml
```diff
@@ -32,5 +33,10 @@ jobs:
...
+      run: python3 ulwgl_test.py
+    - name: Test with unittest
+      if: matrix.python-version == 3.11 || matrix.python-version == 3.12
```
loathingKernel: Same here. Maybe unconditionally run `ulwgl_tests` above for all versions, and run `ulwgl_tests_plugins` here only, for 3.11 and 3.12

loathingKernel:
> We could test that far, but is it fine to use Lutris as our baseline? They must have different requirements to test that far 

Since lutris uses ulwgl i would say yes. We lose nothing doing it anyways. Better be safe.

.github/workflows/ulwgl-python.yml
loathingKernel: Also maybe quote the string of the version number you compare with. Github actions run through js, and sometimes it gets weird.

R1kaB3rN:
> > We could test that far, but is it fine to use Lutris as our baseline? They must have different requirements to test that far
> 
> Since lutris uses ulwgl i would say yes. We lose nothing doing it anyways. Better be safe.

While we have nothing to lose, I don't quite agree with that logic and would prefer that we have more precise reasons for testing that far

Also, if we really want to decisively solve this problem, we have the Flatpak to bundle dependencies that will cover not only Python, but Mesa and Vulkan which DXVK has hard minimum requirements for. But luckily, Jammy sources the _minimum_ Mesa for to use DXVK for gaming to not be a big problem for those users

.github/workflows/ulwgl-python.yml
R1kaB3rN: > I think this should be referenced as `${{ matrix.version }}` for it to work.

It's already working without it

R1kaB3rN: > Also maybe quote the string of the version number you compare with. Github actions run through js, and sometimes it gets weird.

Wow, didn't know that. So this?

> matrix.python-version == "3.10"

But if it's JS, then strictly equals would probably be preferred to avoid any implicit type coercion:

> matrix.python-version === "3.10"

loathingKernel:
> But luckily, Jammy sources the minimum Mesa for to use DXVK for gaming to not be a big problem for those users

Imho the Mesa version requirement for DXVK is irrelevant. Someone might choose to use the OpenGL renderer for whatever reason, or install and use things like gallium nine (I have not tested that option with Proton but you get the point). Someone else might want to chip in here though.

.github/workflows/ulwgl-python.yml
loathingKernel: I had an issue some time ago where I was specifying the matrix like this
```
version: [3.9, 3.10, 3.11, 3.12]
```
instead of
```
version: ['3.9', '3.10', '3.11', '3.12']
```
and for some reason it would lead to interpreting the `3.10` version as `3.1` [link](https://github.com/RareDevs/Rare/actions/runs/7604206484)

I have trust issues towards GH Actions since then and I am always taking the explicit route with them.

R1kaB3rN: To add on, it appears the GH Actions is also strict about quotes as well...

R1kaB3rN: > I think this should be referenced as `${{ matrix.version }}` for it to work.

Sorry, but you were right about this!

PR #41, opened 2024-02-23
Add setup routine for launcher

R1kaB3rN:
On every run of the launcher, we check the state of `~/.local/share/ULWGL` and the configuration file, `ULWGL_VERSION.json`, stored in `/usr/share/ULWGL/ULWGL_VERSION.json` to determine whether an update for the runtime tools or a new install needs to be performed. 

Effectively, `~/.local/share/ULWGL` and `~/.local/share/Steam/compatibilitytools.d` for ULWGL-Runner will always be synchronized with the files in `/usr/share/ULWGL` as determined by the state of the configuration file.

ULWGL_VERSIONS.json:

<pre>
{
  "ulwgl": {
    "versions": {
      "launcher": "0.1-RC3",
      "runner": "0.1-RC3",
      "runtime_platform": "sniper_platform_0.20240125.75305",
      "reaper": "1.0",
      "pressure_vessel": "v0.20240212.0"
    }
  }
}
</pre>

TODO:
- [ ] Change root path to `/usr/share/ULWGL`
- [x] Add tests
- [ ] Support copying Reaper -- (depends on Reaper being included in repo)

### DIFF
--- a/ulwgl_run.py
+++ b/ulwgl_run.py
@@ -11,7 +11,7 @@ from re import match
 from subprocess import run
 from ulwgl_dl_util import get_ulwgl_proton
 from ulwgl_consts import Level
-from ulwgl_util import msg
+from ulwgl_util import msg, setup_ulwgl
 from ulwgl_log import log, console_handler, debug_formatter
 from ulwgl_util import UnixUser
 
@@ -313,11 +313,19 @@ def main() -> int:  # noqa: D103
         "ULWGL_ID": "",
     }
     command: List[str] = []
-    args: Union[Namespace, Tuple[str, List[str]]] = parse_args()
     opts: List[str] = None
+    # Expected files in this dir: pressure vessel, launcher files, runtime platform, runner, config
+    # root: Path = Path("/usr/share/ULWGL")
+    root: Path = Path(__file__).parent
+    # Expects this dir to be in sync with root
+    # On update, files will be selectively updated
+    local: Path = Path.home().joinpath(".local/share/ULWGL")
+    args: Union[Namespace, Tuple[str, List[str]]] = parse_args()
 
     if "ULWGL_LOG" in os.environ:
         set_log()
+
+    setup_ulwgl(root, local)
 
     if isinstance(args, Namespace) and getattr(args, "config", None):
         set_env_toml(env, args)
@@ -353,3 +361,6 @@ if __name__ == "__main__":
     except Exception as e:  # noqa: BLE001
         print_exception(e)
         sys.exit(1)
+    finally:
+        # Cleanup .ref file on every exit
+        Path.home().joinpath(".local/share/ULWGL/.ref").unlink(missing_ok=True)
--- a/ULWGL_VERSION.json
+++ b/ULWGL_VERSION.json
@@ -1,0 +1,11 @@
+{
+  "ulwgl": {
+    "versions": {
+      "launcher": "0.1-RC3",
+      "runner": "0.1-RC3",
+      "runtime_platform": "sniper_platform_0.20240125.75305",
+      "reaper": "1.0",
+      "pressure_vessel": "v0.20240212.0"
+    }
+  }
+}
--- a/ulwgl_consts.py
+++ b/ulwgl_consts.py
@@ -5,6 +5,8 @@ SIMPLE_FORMAT = "%(levelname)s:  %(message)s"
 
 DEBUG_FORMAT = "%(levelname)s [%(module)s.%(funcName)s:%(lineno)s]:%(message)s"
 
+CONFIG = "ULWGL_VERSION.json"
+
 
 class Level(Enum):
     """Represent the Log level values for the logger module."""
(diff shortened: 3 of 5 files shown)


ulwgl_run.py
```diff
@@ -310,3 +318,8 @@ def main() -> int:  # noqa: D103
     except Exception as e:  # noqa: BLE001
         print_exception(e)
         sys.exit(1)
+    finally:
+        # Cleanup .ref file on every exit
+        file: Path = Path.home().joinpath(".local/share/ULWGL").joinpath(".ref")
```
bziemons: Did you know that you can join paths like the following?

```
file: Path = Path.home() / ".local/share/ULWGL" / ".ref"
```

ulwgl_run.py
```diff
@@ -270,12 +270,20 @@ def main() -> int:  # noqa: D103
         "ULWGL_ID": "",
     }
     command: List[str] = []
-    args: Union[Namespace, Tuple[str, List[str]]] = parse_args()
     opts: List[str] = None
+    # Expected files in this dir: pressure vessel, launcher files, runtime platform, runner, config
+    # root: Path = Path("/usr/share/ULWGL")
+    root: Path = Path(__file__).parent
```
bziemons, line 319: This might be too assumptious. The code doesn't check if the path actually is `/usr/share/ULWGL`, but all comments state so. As a packager, I could also imagine chosing `/usr/lib/ULWGL` or lower-case `ulwgl`. Would that be okay?

ulwgl_run.py
loathingKernel: If I might chip in here, I am very strongly against this kind of overloaded operator use. Some operators have very clearly defined usage in other cases, like division here in this case, which makes overloading them essentially code smell.

bziemons: I can respect that it was agreed to not use it

loathingKernel: It didn't come up until now, so there isn't a previous agreement, tbh. I am just stating my personal view on the matter. Imho, if the operator was something more semantically close to the operation it would be fine. @R1kaB3rN is the one doing most of the work developing and maintaining the python code, so the final decision is theirs.

R1kaB3rN: I want the codebase to be as readable as possible. While its a bit more succinct, the result is the same, and sorta reads as traversing a directory from the shell (e.g.` cd ~/.local/share/ULWGL)`, compared to other methods, I believe it can lead to more confusion because the operator conventionally signals a division operation

Unless the efficiency gains, if there is any, outweighs those risks, I can't approve of it

For joining path segments, I believe solutions the involve at least one call to .joinpath()  should be preferred.
Examples:

> Path.home().joinpath(".local/share/ULWGL") # Less preferred because of separators

> Path.home().joinpath(".local", "share", "ULWGL") # Preferred because no separators

>  Path.home().joinpath(".local/" + "share/" + "ULWGL") # Discouraged

>  Path.home().joinpath(".local").joinpath("share").joinpath("ULWGL") # Bad

Currently the codebase should have the less preferred solutions so I'll accept PRs to change them

ulwgl_run.py
R1kaB3rN: Currently, the path for line 276 is used for quick testing purposes as I didn't want to package it to natively in my system and assumes the parent is the ULWGL-Launcher repository. Line 275 will be the real path that will be referred by the launcher once everything is ready. It's just commented at the moment and it doesn't serve as a comment toward line 276. 

As for package management, the launcher (ulwgl-run) is expected to be installed in `/usr/bin` then from there it'll reference the necessary Python modules and other relevant files involved in the setup operation 

cc @GloriousEggroll about paths, package management and to clarify requirements

bziemons: Could you write a bit of documentation into the README or another file about how you expect packaging? As currently, everything resides in `~/.local/share/ULWGL`, I would need to make some adjustments. I already have a PKGBUILD for the current git in AUR ([ulwgl-git](https://aur.archlinux.org/packages/ulwgl-git)). Would look forward to package a released version as well and I would find your input very useful :)

GloriousEggroll: We don't have any current documentation on packaging, however it's already been discussed and planned out within our development team.

```
packages and/or flatpak will do this:

Put everything in /usr/share/ULWGL, with a binary at /usr/bin/ulwgl-run

On first-run, /usr/bin/ulwgl-run will copy the required files from /usr/share/ULWGL to ~/.local/share/ULWGL/ and the steam tool files to ~/.local/share/steam/compatibilitytools.d

After that, any 3rd party tools can just run /usr/bin/ulwgl-run, which from that point forward will only do a version check to make sure the contents of /usr/share/ULWGL/ match that of the one in ~/.local/share/ULWGL
```

GloriousEggroll:
fwiw I've added packaging info to the readme:
https://github.com/Open-Wine-Components/ULWGL-launcher?tab=readme-ov-file#packaging

PR #53, opened 2024-03-03
Add Meson build system

loathingKernel:
Add build system based on Meson. This is not currently working properly and I am opening this PR just for tracking and review purposes.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -47,9 +47,11 @@ version-install: version
 	install -Dm 644 umu/umu_version.json -t $(DESTDIR)$(DATADIR)/$(INSTALLDIR)
 
 
+UMU_RUN_COMMAND = $(DATADIR)/$(INSTALLDIR)/umu_run.py
+
 $(OBJDIR)/.build-umu: | $(OBJDIR)
 	$(info :: Building umu )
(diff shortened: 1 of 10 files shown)


R1kaB3rN:
I just deleted the `flatpak` branch in repo because [update-flatpak](https://github.com/Open-Wine-Components/umu-launcher/tree/flatpak-update) will supersede it. And once that's merged, I'll probably delete that one too.

Though feel free to keep reparenting the branch and reopen if you haven't given up on building with meson.

loathingKernel:
I will rebase on `main` once `update-flatpak` has been merged and re-open an new PR. I don't know if this was closed automatically by GitHub once the target branch was gone, but rebasing and re-targeting branches doesn't require closing the PR normally.

The meson build system is blocked by the `reaper` install location. Meson is being very picky about where to install binaries for subprojects as it is not controlled by umu's meson build system. My workaround would be to include reaper as a subtree instead of a submodule, and adapt it to install in `<prefix>/bin/` as `umu-reaper` instead of `<prefix>/share/umu/reaper` along with the necessary changes in `umu`.

The second point I want to bring up, is that `<prefix>/lib/` would be a more fitting location for the system-wide installation according to FHS. Since `update-flatpak` branch doesn't copy files any more, the change would be much less disruptive now.

What do you think of either of these?

PR #59, opened 2024-03-09
Update documentation

R1kaB3rN:
Adds an initial description section to ulwgl(1) and applies better formatting for the examples/options section.


Aqa-Ib:
I have a question about the env PROTONPATH, please. In practice, what is the difference between not declaring it, and declaring it with its value empty ? Thanks for your work!

PR #66, opened 2024-03-21
Fix fake reaper splitting arguments with spaces

loathingKernel:
Fix the fake reaper splitting the command line arguments due to spaces. Also change the reaper version in ULWGL_VERSION.json to follow the main repo too, since the submodule is already being tracked there. This will cause it to be replaced in case of a package update.


umu/reaper
```diff
@@ -1,5 +1,7 @@
...
 # Fake reaper file to avoid errors while running from source
-args="$*"
-cmd=${args#* -- }
-$cmd
\ No newline at end of file
+while [ "$1" != "--" ]; do
+  shift
+done
+shift
+"$@"
```
R1kaB3rN, line 7: This file should **only** be used in the build process so I think this should return an exit code. That way, the launcher can capture it and handle it accordingly

loathingKernel: The intended use is to strip the reaper part of the command so we can run from source. For me that is useful when debugging through pycharm for example. I cannot see how that related to the build process though.

If you mean to make it just echo the command, print the environement and similar things to aid with debugging without actually running anything, I think that is a good idea and we can do that instead.

If you mean something different, doesn't the shell return the exit code of the last command? I am unsure about it, but if it doesn't, then I'll update it to do so.

R1kaB3rN: Sorry, I was completely misunderstood of the purpose of this file then.

> The intended use is to strip the reaper part of the command so we can run from source. 

No. I'm actually of the opinion that this file should be considered to be permanently deleted to not encourage the practice of running umu from source, even for testing purposes.

But given the severity that this bug can cause, right now, I'll leave that for later or until more feedback that we really want to allow this. But again, I never think we should.

Issue #71, opened 2024-03-26
nix flake


bohanubis:
just here to ask if this project is willing to add nix flake just like neovim for exemple does here: https://github.com/neovim/neovim/tree/master/contrib

I will implement this if its allowed
and thanks

loathingKernel:
I don't think anyone of the contributors so far is using NixOS and personally I don't even know what a nix flake is and what it's meant to provide. Is it a packaging specification?

If so I think it should have a place in this repo, as we already have  an rpm spec file, but you would have to contribute it unless someone else has experience with packaging for NixOS. Of course you will have to make sure it works and that it continues to work by adapting it as the project is still evolving.

bohanubis:
ok gonna make a pull request soon thanks a lot

bohanubis:
@loathingKernel just a quick question what are the architectures this projects support

loathingKernel:
Currently, and for the foreseeable future, x86_64 only, due to Steam runtime.

PR #73, opened 2024-03-27
add nix flake support

bohanubis:
this fixes my issue here: https://github.com/Open-Wine-Components/umu-launcher/issues/71
so what are my questions
am I allowed to add a section about how to use this
basically just something like this
```nix
  inputs = {
    umu= {
      url = "git+https://github.com/Open-Wine-Components/umu-launcher/?dir=packaging\/nix&submodules=1";
      inputs.nixpkgs.follows = "nixpkgs";
    };
 }
{
  environment.systemPackages = [  inputs.umu.packages.${pkgs.system}.umu  ];
}
```
and also possibly a bug since well it will accur in other immutable distros basically now umu copies read only files from /usr/share to .local/share/umu it seems those files can't be modified obviously same with
.local/share/Steam/
so instead of throwing an error
some users will just naivly run it as root
just give the user a message saying
run
chmod -R u+w .local/share/Steam/
chmod -R u+w .local/share/umu/
thats all and thanks

also about that other paricipant well I was stupid enough to put whatever in
"user.name" and "user.email"
so yeah here we are
sorry I have no idea how to fix this mess before creating this


loathingKernel:
> am I allowed to add a section about how to use this
> basically just something like this

You will have to explain what "this" does because I think most of us, myself included, are not familiar with NixOS's packaging. I do not understand what the code snippet is doing.

> and also possibly a bug since well it will accur in other immutable distros basically now umu copies read only files from /usr/share to .local/share/umu it seems those files can't be modified obviously same with

I do not understand how this works. Normally (as in for non-immutable distros) when something is copied as your user into a writable location, the ownership goes to the user doing the copy, is Nix different in this regard? Is there something else that is different about Nix affecting this issue?

>  some users will just naivly run it as root

I agree what we could/should not execute anything if run as root and add a message about it but I don't see how it is connected to the issue above as the issue is unclear to me.

I would like some clarifications about these things, it can help us figure out what should be handled by umu itself and what should be part of packaging. Also feel free to explain them in detail as I think most people are not familiar with the intricacies of NixOS.

R1kaB3rN:
> ```nix
>   inputs = {
>     umu= {
>       url = "git+https://github.com/Open-Wine-Components/umu-launcher/?dir=contrib\/nix&submodules=1";
>       inputs.nixpkgs.follows = "nixpkgs";
>     };
>  }
> {
>   environment.systemPackages = [  inputs.umu.packages.${pkgs.system}.umu  ];
> }
> ```
Thanks for this contribution, and I'm very open to having a Flake included. I'll have to look into how it works in a bit though, and I don't think anyone from the team is very familiar with NixOS either.

> and also possibly a bug since well it will accur in other immutable distros basically now umu copies read only files from /usr/share to .local/share/umu it seems those files can't be modified obviously same with .local/share/Steam/ so instead of throwing an error some users will just naivly run it as root just give the user a message saying run chmod -R u+w .local/share/Steam/ chmod -R u+w .local/share/umu/ thats all and thanks

Yes, the launcher (umu_run.py) and the wrapper script (umu-run) should easily be able to handle this case -- not running as root.

Also, are file permissions not able to be set in the Flake file? The launcher copies files via [copy](https://docs.python.org/3/library/shutil.html#shutil.copy) which respects the file permissions mode, so if permissions are properly setup during packaging phase then we can avoid having to display that message.

bohanubis:
so I'll make two comments first one explaining what each file does since well why not
and the second one will contain that problem with running as root and write permissions

I'll start with the first comment
so to install this flake the user should add
```nix
    umu= {
      url = "git+https://github.com/beh-10257/umu-launcher-nix/?dir=packaging\/nix&submodules=1";
      inputs.nixpkgs.follows = "nixpkgs";
    };
```
disclaimer (I am using my repo since this is not merged yet)
to his inputs
so what does that mean it tells nix to use git to clone this repo https://github.com/beh-10257/umu-launcher-nix somewhere in /nix/store/
`?dir=packaging\/nix` tells that the flake.nix exists in `packaging/nix` directory and submodules=1 tells it to download git submodules
aka: https://github.com/Open-Wine-Components/umu-launcher/tree/main/subprojects
since we don't want an empty directory there
`inputs.nixpkgs.follows = "nixpkgs";` you can think of nixpkgs as the repos of a distro this says to nix to ignore the version I provided and use the version in the user config.
this is just so the user doesn't have two version of nixpkgs in their system the version in their config and the one defined in flake.lock here

and next they should add this somewhere in their flake
```nix
{ inputs, ... }:
{
  environment.systemPackages = [
    inputs.umu.packages.${pkgs.system}.umu
  ];
}
```
it basically goes to the umu input and takes the umu package \\ ${pkgs.system} translates to x86_64-linux aka the architecture of the package

------
next I'll explain what the files commited to this repo do
first flake.nix
```nix
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
  };
```
this says to the flake to use the unstable version of nixpkgs
the specific commit of nixpkgs is fixed if flake.lock
just making a quick look at flake.lock pretty much explains everything
image
for exemple it fixes the revision so without adding this line `inputs.nixpkgs.follows = "nixpkgs";` in the user config
as was mentioned before
it will download nixpkgs with this commit
"57e6b3a9e4ebec5aa121188301f04a6b8c354c9b"

---
next in flake.nix we have this line
`umu-package = nixpkgs.legacyPackages.x86_64-linux.callPackage ./umu-launcher.nix { umu-launcher=( builtins.toPath "${self}/../../"); };`
this defines a variable (the type is a package btw; yes paths and packages are actual variable type in nix language)
anyway this basically just calls umu-launcher.nix for the package definition and passes along some variables
umu-launcher is just the name of the path to the source of this repo aka: https://github.com/Open-Wine-Components/umu-launcher
it will be used in umu-launcher.nix file obv
builtins.toPath "${self}/../../" basically is the path to that source in the /nix/store as I said in the beginning
${self} returns where is flake.nix specifically
heres an example
image
in this exemple ${self} is "/nix/store/2kgj382l1931vdjf5vi6rrpycs6n6sxy-source/packaging/nix"
thats why we should add ../../
also toPath since well we should transform strings to paths if we want to build the thing
anyway this all passes along to umu-launcher.nix

---
```nix
{stdenv , umu-launcher, pkgs, ...}:
stdenv.mkDerivation {
  pname = "umu-launcher";
  version = "0.1";
  src = umu-launcher;
  depsBuildBuild = [
    pkgs.meson
    pkgs.ninja
    pkgs.scdoc
  ];
  dontUseMesonConfigure = true;
  dontUseNinjaBuild = true;
  dontUseNinjaInstall = true;
  dontUseNinjaCheck = true;
  configureScript = "./configure.sh";
}
```

> so as you can see its taking the aforementioned umu-launcher variable
stdenv.mkDerivation is basically the regular build enviroment in nix to use make and that kinda stuff (what this project is using)
src is umu-launcher aka the path to the repo aka what was said before (toPath and stuff)
```nix
  dontUseMesonConfigure = true;
  dontUseNinjaBuild = true;
  dontUseNinjaInstall = true;
  dontUseNinjaCheck = true;
  configureScript = "./configure.sh";
```
all other variables are self explanatory right ?? (basically nix is stupid and I have to tell him not to do stuff automatically)
this all now provides
image
which is great but when I run this in an actual game
we get this error
image
```
NixOS cannot run dynamically linked executables intended for generic
linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld
```
anyway yeah that link brings me here: https://nix.dev/guides/faq#how-to-run-non-nix-executables
basically generic libraries can't be run in nix beside the fact it doesn't follow FHS standard
image
I guess the executable in question is
image
so yeah I followed that page to create an FHS like enviroment

thats the next variable in flake.nix, another package and I named the variable umu-run
and it take the original package as an input and is defined in umu-run.nix
so between let in I can define variables and well
```nix
  ldPath = lib.optionals stdenv.is64bit [ "/lib64" ] ++ [ "/lib32" ];
  exportLDPath = ''
    export LD_LIBRARY_PATH=${lib.concatStringsSep ":" ldPath}''${LD_LIBRARY_PATH:+:}$LD_LIBRARY_PATH
  '';
```
this basically adds /lib64:/lib32: to the start of that variable (this is basically stolen on how steam is packaged in nix)
but I mean it makes sense in my head
image
except that fact I just did this as well
image
to see the difference
```
/run/opengl-driver/lib:/run/opengl-driver-32/lib:/etc/sane-libs:/nix/store/cvalskd4xbqf5f7srlzl60213s0s3fwz-pipewire-1.0.4-jack/lib
after exporting
/lib64:/lib32:/run/opengl-driver/lib:/run/opengl-driver-32/lib:/etc/sane-libs:/nix/store/cvalskd4xbqf5f7srlzl60213s0s3fwz-pipewire-1.0.4-jack/lib
```
anyway next `name="umu"` umu is what should the user run to be in the enviroment and not have `NixOS cannot run dynamically linked executables intended for generic linux environments`

---
next 
```nix
  runScript = writeShellScript "umu-env" ''
    ${exportLDPath}
    ${package}/bin/umu-run "$@"
  '';
```
the ${package} here is the path to umu-package aka
image
"/bin/umu-run" goes to the binary as you can image and "$@" passes whatever is after umu
aka here
`umu path_to_exe` "$@" is replaced with path_to_exe I guess
so now we have a working command `umu` which runs umu correctly

---
anyway now the problem we have is this
image
in our Path
this is confusing and not helpful
so we should find a solution we can do this in flake.nix to fix the problem
```nix
{
  description = "umu universal game launcher";
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    umu-launcher = {
      flake = false;
      url="git+https://github.com/Open-Wine-Components/umu-launcher?submodules=1";
    };
  };
  outputs = { self, nixpkgs, umu-launcher }:
  let
    umu-package = nixpkgs.legacyPackages.x86_64-linux.callPackage ./umu-launcher.nix { umu-launcher=umu-launcher; };
  in
  let
    umu-run = nixpkgs.legacyPackages.x86_64-linux.callPackage ./umu-run.nix { package=umu-package; };
  in{
    packages.x86_64-linux.umu-run = umu-run;
    #packages.x86_64-linux.umu-package = umu-package;
  };
}
```
this will basically remove umu-run from PATH but also `man umu` will not work since well man page comes from umu-package and that line is now commented out which is a problem since I want man pages to work

---
basically
we want the binary from umu-run and the man pages from umu-package
btw all other functionality works
since umu-package is a dependency of umu-run so everything is there its just not exposed so we should expose it
so yeah there is obviously a solution now to the final part
I'm defining a package called umu package (the only exposed output in the commited flake.nix)
its as always talking combine.nix and taking two variables 2 packages the environment and the actual package
combine.nix is obvious I think like read it anyway
the final output is a working umu and man page
image
so yeah great
I think I explained everything
yay it works
image

loathingKernel:
Thank you for the very detailed description, personally I will need to read it a few times to comprehend it. In the meantime, one easy to fix issue that I missed before is that we already use the `packaging` directory for packaging related stuff. Please use that instead of `contrib` to keep things tidy.

bohanubis:
anyway now running as root and the read only problem
as you can see first of all
image
/nix/store directory is read only aka immutable distro and remember basically everything related to system (including packages including umu) is symlinked to /nix/store
also 
image
and
heres the actual problem
image
image
so
image
yay this works now
anyway as far as I'm concerned we can just make a disclaimer in the installation section about immutable distros
and the least we can do is to never allow users to use umu with sudo

bohanubis:
@loathingKernel done

loathingKernel:
So if I understand correctly the issue is that Nix, doesn't have the `writable` bit for `user`. @R1kaB3rN maybe this can simply be fixed by using `os.chmod()` to set the correct bits on the user-local files after being copied. All we need is `write` for `user` for the whole `.local/share/umu` directory, preferably through a mask to retain the `execute` permission on each file. This could be useful to other distros too.

bohanubis:
so should I edit
https://github.com/Open-Wine-Components/umu-launcher/blob/main/README.md

if I will at which section should I add this
as far as I can see there is no specific section for installing for distros
can I just through it in installing section or something
or maybe in a new section where there will be nix flatpak in that section aka how to install

loathingKernel:
IMHO, a new section after `Building`, titled `Packages` or something similar would be the better approach for this, and a subsection for NixOS. We can add more later as they become available.

R1kaB3rN:
There's many ways to go about it in the README. If it were me, I'd combine the _Installing_ and _Installing as user_ section to a _make_ section then have the next section be NixOS.

For example:
### Install
#### make
#### NixOS

README.md
```diff
@@ -110,6 +110,26 @@ You will need to add `$HOME/.local/bin` in your `$PATH` to be able to run umu th
 export PATH="$HOME/.local/bin:$PATH"
 ```
 
+# Packages:  
+## NisOS
```
loathingKernel: Typo

README.md
```diff
@@ -110,6 +110,26 @@ You will need to add `$HOME/.local/bin` in your `$PATH` to be able to run umu th
...
+if you want to add umu-launcher as a flake add this to your inputs in flake.nix
+```nix
+  inputs = {
+    umu= {
+      url = "git+https://github.com/Open-Wine-Components/umu-launcher/?dir=packaging\/nix&submodules=1";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  }
+```
+and in your configuration.nix (or obv any file in your config)
```
loathingKernel: obv -> obviously

It is a Readme, needs some formality or at least not abbreviated words. Someone might try translate it through an automated tool. IMHO this needs to be rephrased.

loathingKernel (changes requested):

README.md
bohanubis: done

README.md
loathingKernel: you do not need to have the first letter as capital. Auto-correct did it for me. lemme fix that

bohanubis: I fixed this I will keep this thread open though
for well (I'll probably think of a suggestion)
and sorry for not responding right away I was messing around with git branches
and I messed up lol https://github.com/Open-Wine-Components/umu-launcher/pull/77
anyway I'll think of better wording
and FWIW in nix you can really put that wherever you want

bohanubis: @loathingKernel done

loathingKernel: >  and FWIW in nix you can really put that wherever you want

From experience, users have a tendency to misinterpret instructions in very creative ways. I think that just mentioning the proper/intended/documented way of doing it is enough. Users that know where they can deviate from instructions with alternative ways to do something, will use them despite the documentation. For the rest of them, only being aware of the "proper" way helps with consistency and reproducibility.

bohanubis: @loathingKernel than I guess I will just remove that obvious section now
if you think this is enough I guess resolve this thread
and this is ready to be merged probably

loathingKernel: What it should say is up to you, you are the knowledgeable person with how NixOS works. I am merely giving suggestions on how instructions should be structured and formatted. You have to figure out what is relevant and what isn't in this case because it is unlikely these instructions can be later edited by us while retaining your intended information. For me, once **you** are happy with the instructions, this conversation can be marked as resolved.

bohanubis: FWIW I used very generic names most people in the nix sphere will know what they are doing so yeah the state the nixos section is in now is good enough to be honest

and also to be honest I'm sure this will not be the final stage of that section so yeah this can be resolved

loathingKernel (approved):
Looks good to me content-wise. Only nitpick is that since #75 is merged in `main` now, if it would be possible to rebase the changes on it, it would result in a cleaner hsitory. This isn't a blocker though as we could also squash-merge, it's up to you.

bohanubis:
@loathingKernel I mean I think I already did so
in here: https://github.com/beh-10257/umu-launcher-nix/commits/main/
image

image
(thats why I was messing around with git rebase in the first place)
also the reason they are 24 commits is well I managed to somehow have a couple of the same commits committed at the same time
I have no idea why is that and I am so sorry for messing it up (I am very new to using git so yeah)

R1kaB3rN:
@beh-10257 Before I merge this, to clarify again, you weren't able to get man pages to work?

R1kaB3rN:
@beh-10257 Thanks for your contribution! I'll be closing this as its been merged at https://github.com/Open-Wine-Components/umu-launcher/commit/81707ad2bc0fb648c111e46f635ebe87bd79516c

bohanubis:
@R1kaB3rN no
man pages are working properly
its just that the executable name is `umu` instead of `umu-run`
and man `umu` works perfectly fine
sorry for not explaining well enough

bohanubis:
@R1kaB3rN you forgot about this I supposed https://github.com/Open-Wine-Components/umu-launcher/pull/73/commits/e85a921956479e8fb946f2917c95af2411bd8b06

I checked already done sorry sorry

bohanubis:
basically this
image

R1kaB3rN:
> @R1kaB3rN you forgot about this I supposed [e85a921](https://github.com/Open-Wine-Components/umu-launcher/commit/e85a921956479e8fb946f2917c95af2411bd8b06)
> 

Fixed in https://github.com/Open-Wine-Components/umu-launcher/commit/237d574eed88ef401e00376824f1d555cee3a495

PR #75, opened 2024-03-28
Add support for immutable distributions

R1kaB3rN:
Related to https://github.com/Open-Wine-Components/umu-launcher/pull/73 and https://github.com/Open-Wine-Components/umu-launcher/issues/71

Ensures that write permissions will be added to the destination files when copying from a read-only filesystem.

### DIFF
--- a/umu/umu_consts.py
+++ b/umu/umu_consts.py
@@ -35,3 +35,10 @@ PROTON_VERBS = {
     "getcompatpath",
     "getnativepath",
 }
+
+
+class MODE(Enum):
+    """Represent the permission to apply to a file."""
+
+    USER_RW = 0o0644
+    USER_RWX = 0o0755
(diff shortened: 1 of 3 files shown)


R1kaB3rN:
cc @beh-10257

bohanubis:
@R1kaB3rN so I tested this
it works great it even keeps permission for group and everyone which is great
image
image

for context I ran the first `ls -lah` before rebuilding umu with this pull request
so I nuked everything before seeing what will happen it worked perfectly so yeah this works
thanks for working on this

R1kaB3rN:
> for context I ran the first `ls -lah` before rebuilding umu with this pull request so I nuked everything before seeing what will happen it worked perfectly so yeah this works thanks for working on this

I just pushed some new changes and plan to merge this in main. If you can take the time to retest it again, that'd be great. If you encounter any issue again, please feel create another report.

Also, one of your environment variables is incorrect -- it's actually `UMU_LOG` not `umu_LOG`. The environment variable's name got messed up after the name change, which is why you're not seeing any logs at all.

@beh-10257

bohanubis:
@R1kaB3rN if I can fix the man pages I will

and yeah I tested the change and they work great

image
image

PR #80, opened 2024-04-01
Flatpak and makefile

loathingKernel:
Minor changes for flatpak

cc @R1kaB3rN

PR #82, opened 2024-04-02
Update Flatpak support

R1kaB3rN:
This pull request updates the Flatpak manifest and the launcher to support running in a Flatpak environment. Flatpak clients who wish to use umu as their backend will be required to change their manifest to permit access to umu's $XDG_DATA_HOME/umu directory where the Steam Runtime platform and the launcher files will be stored/copied to.

### DIFF
--- a/packaging/flatpak/org.openwinecomponents.umu.launcher.yml
+++ b/packaging/flatpak/org.openwinecomponents.umu.launcher.yml
@@ -60,7 +60,6 @@ finish-args:
   - --share=network
   - --talk-name=org.freedesktop.Notifications
   - --talk-name=org.kde.StatusNotifierWatcher
-  - --talk-name=org.freedesktop.Flatpak
   # Required for bwrap to work
   - --talk-name=org.freedesktop.portal.Background
   # --- Steam ---
--- a/packaging/flatpak/org.openwinecomponents.umu.launcher.yml
+++ b/packaging/flatpak/org.openwinecomponents.umu.launcher.yml
@@ -60,8 +60,13 @@ finish-args:
   - --share=network
   - --talk-name=org.freedesktop.Notifications
   - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Flatpak
   # Required for bwrap to work
   - --talk-name=org.freedesktop.portal.Background
+  # --- Steam ---
+  # Pressure Vessel
+  # See https://github.com/flathub/com.valvesoftware.Steam/commit/0538256facdb0837c33232bc65a9195a8a5bc750
+  - --env=XDG_DATA_DIRS=/app/share:/usr/lib/extensions/vulkan/share:/usr/share:/usr/share/runtime/share:/run/host/user-share:/run/host/share:/usr/lib/pressure-vessel/overrides/share
 
 add-extensions:
   org.freedesktop.Platform.Compat.i386:
@@ -143,14 +148,14 @@ modules:
   - name: umu-run
     buildsystem: simple
     build-commands:
-      - install -D umu-run-cli /app/bin/umu-run
-      - install -D umu-launcher.tar.gz /app/share/umu/umu-launcher.tar.gz
+      - |
+        git submodule update --init --recursive
+        ./configure --prefix=/app
+        make install
     sources:
-      - type: file
-        path: umu-run-cli
-      - type: file
-        url: https://github.com/Open-Wine-Components/umu-launcher/releases/download/0.1-RC3/umu-launcher.tar.gz
-        sha256: e25c4dd0636d04e7c8c534cf3c5bbdca5ae0d49f146ee8395306174700899952
+      - type: git
+        url: https://github.com/Open-Wine-Components/umu-launcher.git
+        branch: main
 
   - name: platform-bootstrap
     buildsystem: simple
--- a/packaging/flatpak/org.openwinecomponents.umu.launcher.yml
+++ b/packaging/flatpak/org.openwinecomponents.umu.launcher.yml
@@ -150,7 +150,7 @@ modules:
     build-commands:
       - |
         git submodule update --init --recursive
-        ./configure --prefix=/app
+        ./configure.sh --prefix=/app
         make install
     sources:
       - type: git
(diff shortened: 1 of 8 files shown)


R1kaB3rN:
cc @GloriousEggroll

packaging/flatpak/org.openwinecomponents.umu.launcher.yml
```diff
@@ -143,14 +148,14 @@ modules:
...
+        make install
     sources:
-      - type: file
-        path: umu-run-cli
-      - type: file
-        url: https://github.com/Open-Wine-Components/umu-launcher/releases/download/0.1-RC3/umu-launcher.tar.gz
-        sha256: e25c4dd0636d04e7c8c534cf3c5bbdca5ae0d49f146ee8395306174700899952
+      - type: git
+        url: https://github.com/Open-Wine-Components/umu-launcher.git
+        branch: main
```
R1kaB3rN: This won't build since the changes are not merged in main, so will probably have to create another branch

umu/umu_consts.py
```diff
@@ -36,9 +40,12 @@ class Color(Enum):
...
-    """Represent the permission to apply to a file."""
+FLATPAK_PATH: Path = Path(environ.get("XDG_DATA_HOME"), "umu") if FLATPAK_ID else None
 
-    USER_RW = 0o0644
-    USER_RWX = 0o0755
+UMU_LOCAL: Path = (
+    FLATPAK_PATH if FLATPAK_PATH else Path.home().joinpath(".local", "share", "umu")
+)
+
+UMU_CACHE: Path = Path.home().joinpath(".cache", "umu")
```
loathingKernel: Which `.cache` is this? Flatpak's or hosts? what is the intended behavior here?

umu/umu_run.py
```diff
@@ -239,6 +251,9 @@ def build_command(
         err: str = "The following file was not found in PROTONPATH: proton"
         raise FileNotFoundError(err)
 
+    if FLATPAK_PATH and Path(__file__).resolve().parent == Path("/app/share"):
```
loathingKernel: Why both clauses? what does the second one achieve? Is it required? Seems like it is hardcoding a path which can be bad if it isn't explicitly required or can work without.

umu/umu_run.py
```diff
@@ -304,6 +319,13 @@ def main() -> int:  # noqa: D103
 
     log.debug("Arguments: %s", args)
 
+    if FLATPAK_PATH and root == Path("/app/share"):
```
loathingKernel: Same question as above.

packaging/flatpak/org.openwinecomponents.umu.launcher.yml
```diff
@@ -143,14 +148,14 @@ modules:
   - name: umu-run
     buildsystem: simple
     build-commands:
-      - install -D umu-run-cli /app/bin/umu-run
-      - install -D umu-launcher.tar.gz /app/share/umu/umu-launcher.tar.gz
+      - |
+        git submodule update --init --recursive
+        ./configure --prefix=/app
```
loathingKernel: it is `configure.sh` not `configure`

loathingKernel (commented):
* If we decide to spawn SLR through `flatpak-spawn` can you also make the flatpak manifest simpler?
* Also we should update the the makefile to create a special `umu-run` for umu-launcher's Steam integration, to call the flatpak instead of `.local/share/umu/umu_run.py`, so we can make sure we always run the flatpak to update the user-local files on umu updates.

umu/umu_run.py
R1kaB3rN: There's actually a typo there -- it's actually /app/share/umu. 

I want to check for the entry point to ensure that the Flatpak application is being launched instead of just trusting the path/environment variable inputs.

loathingKernel: I still find it superfluous, if the user wants to mess with variables or paths, this won't stop them, but it's not important enough to  argue over it. I am fine with it keeping it as is.

packaging/flatpak/org.openwinecomponents.umu.launcher.yml
```diff
@@ -143,14 +148,14 @@ modules:
   - name: umu-run
     buildsystem: simple
     build-commands:
-      - install -D umu-run-cli /app/bin/umu-run
-      - install -D umu-launcher.tar.gz /app/share/umu/umu-launcher.tar.gz
+      - |
+        git submodule update --init --recursive
+        ./configure.sh --prefix=/app
+        make install
```
loathingKernel: The required `FLATPAK=xtrue` control variable to disable manpages is missing here. For reference: https://github.com/Open-Wine-Components/umu-launcher/pull/80/files#diff-04d351e49f21347db94177694691013189c0c7e751a2894b523df166b3b5b5c8R154

loathingKernel:
> Also we should update the the makefile to create a special umu-run for umu-launcher's Steam integration, to call the flatpak instead of .local/share/umu/umu_run.py, so we can make sure we always run the flatpak to update the user-local files on umu updates.

Attaching a patch for the makefile to resolve this [0001-makefile-Install-umu-run-for-umu-launcher-steam-tool.patch.txt](https://github.com/Open-Wine-Components/umu-launcher/files/14848456/0001-makefile-Install-umu-run-for-umu-launcher-steam-tool.patch.txt)

Patch does not include the required changes in the python code to not symlink `umu_run.py` into the steam compatibility tool.

loathingKernel:
Given that we use a script for flatpak in steam's compat tool, using symlinks for the system packages is of no value and just convolutes the code. Consider using a script in that case too. The makefile is already creating it and installing it for the package.

R1kaB3rN:
> Given that we use a script for flatpak in steam's compat tool, using symlinks for the system packages is of no value and just convolutes the code. Consider using a script in that case too. The makefile is already creating it and installing it for the package.

It's fixed, but because there was a bug since the launcher would create the symbolic link to the entry point when it had already been created during the build process, resulting in a FileExistsError. This was a failure on my part for not testing it.

> using symlinks for the system packages is of no value and just convolutes the code.

Let's please be careful not to reopen our past disagreement about symbolic links, whether that's your intention or not. Because to be honest, I'm still not quite settled or convinced about it and I only let that pass solely because I value your contribution to this project and would like to encourage others to contribute as well. However, if there's any complaints about it from anyone, causes any kind of issue or bugs like [you did here](https://github.com/Open-Wine-Components/umu-launcher/pull/66), which I doubt, I may have to reconsider it.

PR #109, opened 2024-06-05
Add functionality to use winetricks

R1kaB3rN:
Add initial support for executing winetricks without writing a protonfix module or using the winetricks GUI via GAMEID=winetricks-gui.

Example Usages:

> GAMEID=0 umu-run winetricks quartz
> GAMEID=0 umu-run winetricks quartz wmp11 qasf

Edit: Changing the layout of the filesystem via bwrap to use the system winetricks needs more discussion and exploration. This case is only relevant when using Valve's Proton builds, and it'll require having Steam running in the background and writing a file that contains an app ID in a folder. Of course, this is not good as it will probably require installing Steam which is contrary to goals of this project. GE-Proton and UMU-Proton builds are not effected by this, and have patches that solve those problems. Launchers such as Heroic or Lutris can workaround this issue but for umu, unless there's a way to workaround those two problems without installing Steam, this feature will need to wait.

### DIFF
--- a/docs/umu.1.scd
+++ b/docs/umu.1.scd
@@ -118,6 +118,12 @@ $ WINEPREFIX=~/.wine GAMEID=0 PROTONPATH=GE-Proton9-1 umu-run ~/foo.exe
 $ WINEPREFIX=~/.wine GAMEID=0 PROTONPATH=GE-Proton umu-run ~/foo.exe
 ```
 
+*Example 11. Install winetricks verbs*
+
+```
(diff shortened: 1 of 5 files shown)


loathingKernel (commented):
Solely as an FYI, stable versions of Proton up to [Proton 8 used different paths](https://github.com/ValveSoftware/Proton/blob/proton_8.0/proton#L411-L426)

I don't think this is something that should be accounted for necessarily, but it is good to know if someone files an issue at some point trying to run winetricks with an older Proton version.

PR #126, opened 2024-06-23
Replace PurePath.as_posix with f-string expressions

R1kaB3rN:
The launcher interacts with filesystem paths via `pathlib` and uses `PurePath.as_posix` to convert the object back to a string to set the string value to an appropriate environment variable. However, using `PurePath.as_posix` is inefficient. Please see commit message for details. 

Instead of `PurePath.as_posix`, the launcher can simply use f-strings which is more efficient and succinct.

### DIFF
--- a/umu/umu_plugins.py
+++ b/umu/umu_plugins.py
@@ -27,9 +27,7 @@ def set_env_toml(
         raise ModuleNotFoundError(err)
 
     toml: dict[str, Any] = None
-    path_config: str = (
-        Path(getattr(args, "config", None)).expanduser().as_posix()
-    )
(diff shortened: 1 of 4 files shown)


loathingKernel:
Why use fstrings instead of calling `str()` directly?

R1kaB3rN:
> Why use fstrings instead of calling `str()` directly?

Mainly for consistency since the codebase already uses f-strings in a few places, but I would actually be fine with using `str()`  and might prefer it more in a certain places. 

I think using str() would be better taste in cases for single expressions and would probably be considered more Pythonic.

Issue #63, opened 2024-03-12
Launching a game which executable is in a subdirectory of the main directory


Aqa-Ib:
Imagine the following folder structure for a game:
- Game main folder: /home/user/game
- Executable folder: /home/user/game/x64/launcher.exe

In this case, the main game directory is different than the executable directory.

This game won't work when launched like this:
- `cd /home/user/game/`
- `ulwgl-run /home/user/game/x64/launcher.exe`

Using vanilla wine, a solution is to run it like this: 
- `cd /home/user/game/x64/`
- `wine start /d "/home/user/game" launcher.exe`

Using Steam, we have the "Start in" field available:

![image](https://github.com/Open-Wine-Components/ULWGL-launcher/assets/16420574/fe32682c-525e-444f-9c3d-6df599d272c4)

Using Lutris we have a "Working directory" field.

How do we achieve this using `ulwgl-run` ?

Aqa-Ib:
Alright I got it, it works when launched with a relative path to the current working directory:

- `cd /home/user/game/`
- `ulwgl-run /x64/launcher.exe`

_Notice that the ulwgl-run command can't contain the full path in this case or it won't work._

Aqa-Ib:
It would be nice if the working directory could be defined into the ulwgl-run command though, like Steam and Lutris do.

R1kaB3rN:
What's your specific use case? Personally, I've never used it when launching my games so I never saw the point of changing to a specific directory before launching a game.

I can give you this feature if it's not too niche, but it'd be via the configuration file usage. For example:

> [ulwgl]
> exe = "/home/user/game/x64/launcher.exe"  
> start = "/home/user/game/" # Directory to change to
> proton = "/home/user/.local/share/Steam/compatibilitytools.d/GE-Proton9-1"
> prefix = "/home/user/.wine"
> game_id = "0"

Aqa-Ib:
For some games I need to give ulwgl-run the .exe path relative to the cwd (current working directory).

When the game has its executable in a directory below the main game directory, it won't work without a relative path to the cwd. 

However, I have also found a bug with ulwgl-run that is not present in vanilla wine. Consider a game directory with the main executable called frontend.exe, and a secondary executable called start.exe.

- `cd /home/user/game`
- If I execute `ulwgl-run frontend.exe` it works.
- If I execute `ulwgl-run /home/user/game/frontend.exe` it will launch `start.exe` instead.

[...]

PR #138, opened 2024-07-04
Attempt to find game working directory

R1kaB3rN:
Related to https://github.com/Open-Wine-Components/umu-launcher/issues/63


Aqa-Ib:
Excuse me, but I don't understand how may I execute the executable contained in the subdir considering this patch ?

R1kaB3rN:
> Excuse me, but I don't understand how may I execute the executable contained in the subdir considering this patch ?

In the patch at that time, all you needed to do was simply run the GOG game by passing the absolute path then umu would change to the correct directory for you. As a result, you wouldn't need to worry about changing to any directory in your helper script or before running umu from the CLI.

And long story short, this feature has been put on hold.

<details>
<summary>Explanation</summary>
<p>After some feedback from the team, it was determined that it wouldn't really be too useful for clients who are already in a position to change to the correct directory and it risks conflicting with the user's target directory set in the client. Since then, the direction of the patch has changed to only finding the correct directory for Steam games but that has also been put on hold too due to the uncertainty of whether to even support that use case.</p>

<p>If this feature were to be added though, it'd probably be in the configuration file usage as it wouldn't get in the way of client apps, thus posing no risk. Also, it'd need to be really useful to those users.</p>
</details>

PR #140, opened 2024-07-07
Install UMU to be in Python site-packages

R1kaB3rN:
Changes umu to be installed as an appropriate Python module, which is in the user's Python site-packages directory instead of `/usr/share/umu` and allows umu clients to specify their runtime path to find its dependencies (e.g., python-xlib).

As a result, the following new build tools will be required: [hatchling](https://github.com/pypa/hatch), [installer](https://github.com/pypa/installer) and [build](https://github.com/pypa/build), and clients will be expected to add umu's dependencies within the directory where umu is installed or by passing a path to `UMU_CLIENT_RTPATH`. For Heroic users, this means having python-xlib within `$HOME/.config/heroic/tools/runtimes/umu`


alterNERDtive:
This breaks installing as user instead of system-wide: https://github.com/Open-Wine-Components/umu-launcher#installing-as-user

````bash
> ./configure.sh --user-install
:: Created ./Makefile, now run "make" to build.
> make install
:: Installing umu_version.json
install -d /usr/local/lib64/python3.12/site-packages/umu-launcher
install: cannot create directory ‘/usr/local/lib64/python3.12’: Permission denied
make: *** [Makefile.in:46: version-install] Error 1
````

PR #144, opened 2024-07-11
fixes nix build issue and updates nixpkgs

bohanubis:
I also got ahead and make the maintainers bit the first thing in the nixos section
anyway this should fix the issue
I don't know if my commit missages are good enough or not so yeah
I tried to use fixes this time lol

also I have no idea why its `Unverified` so yeah

### DIFF
--- a/packaging/nix/umu-launcher.nix
+++ b/packaging/nix/umu-launcher.nix
@@ -1,19 +1,35 @@
-{stdenv , umu-launcher, pkgs, ...}:
-stdenv.mkDerivation {
-  pname = "umu-launcher";
+{pyth1 ,python3Packages , umu-launcher, pkgs, ...}:
+python3Packages.buildPythonPackage {
+  name = "umu-launcher";
   version = "0.1";
   src = umu-launcher;
+  pyproject = false;
   depsBuildBuild = [
     pkgs.meson
     pkgs.ninja
     pkgs.scdoc
+    pkgs.git
+    pkgs.python3Packages.installer
+    pkgs.hatch
+    pkgs.python3Packages.build
   ];
   propagatedBuildInputs = [
-    pkgs.python3
+    pyth1
+    pkgs.python3Packages.xlib
+    pkgs.gvfs
+    pkgs.dconf
   ];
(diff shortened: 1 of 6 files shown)


packaging/nix/umu-launcher.nix
```diff
@@ -1,19 +1,33 @@
...
-stdenv.mkDerivation {
-  pname = "umu-launcher";
+{pyth1 ,python3Packages , umu-launcher, pkgs, ...}:
+python3Packages.buildPythonPackage {
+  name = "umu-launcher";
   version = "0.1";
   src = umu-launcher;
+  pyproject = false;
   depsBuildBuild = [
     pkgs.meson
```
R1kaB3rN, line 8: You can safely remove `meson` as a build dependency

R1kaB3rN: Same with `ninja` as well, since we're no longer building `reaper`

R1kaB3rN:
Also, in the future, it would be nice to have the Nix package built in the CI just like the Ubuntu package.

Going to merge this for now because I want to fix that issue so Nix users can have something working. Removing of `ninja` and `meson` as a build dependency is minor and can be addressed later.

bohanubis:
pyproject = false;
  this is needed since there is not setup.py in this project
  I just checked there is no need as you said for meson and ninja
  basically just remove the
```
  pkgs.meson
  pkgs.ninja
```
  line
  should I open another pull request ??
  also
  `Nix package built in the CI just like the Ubuntu package.`
  by this you mean like a test so when a commit happen the CI can tell when the nix package breaks
  is that what you mean ??

R1kaB3rN:
> pyproject = false; this is needed since there is not setup.py in this project I just checked there is no need as you said for meson and ninja basically just remove the
> 
> ```
>   pkgs.meson
>   pkgs.ninja
> ```
> 
> line should I open another pull request ?? also `Nix package built in the CI just like the Ubuntu package.` by this you mean like a test so when a commit happen the CI can tell when the nix package breaks is that what you mean ??

Yes, when you can make time, feel free to open another pull request to update the flake. And yes, that is what I mean. I think it would be _very_ helpful and a nice-to-have if we can build the Nix package in a Github Action.

bohanubis:
yeah I will need to learn how to write github actions I guess

PR #149, opened 2024-07-15
make rule to generate standalone python zippapp application

db47h:
This adds a `zipapp` rule that bundles umu-run and its dependencies in a standalone executable (a python zipapp), ideal for clients like heroic or lutris.

This can also be used from python code by adding the file to PYTHONPATH (or sys.path) as in:

PYTHONPATH=~/.local/bin/umu-run python -m umu

No additional build deps are required. I could add a config variable to exclude dependencies for clients that manage that themselves.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -178,4 +178,14 @@ user-install:
 	$(info :: To run you need to make sure "$(BINDIR)" is in your PATH )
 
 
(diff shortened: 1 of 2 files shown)


R1kaB3rN:
Closing as it will be superseded by https://github.com/Open-Wine-Components/umu-launcher/pull/174

PR #154, opened 2024-07-19
Move handling of umu_version.json to pyproject.toml

db47h:
New build dependency: hatch vcs

Removed the version* targets from the Makefile.

The generated version is more convoluted than before though: `"launcher": "0.1rc5.dev234+g9cbd15d.d20240719"`

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -22,9 +22,9 @@ FLATPAK ?= xfalse
 
 .PHONY: all
 ifeq ($(FLATPAK), xtrue)
-all: version umu umu-launcher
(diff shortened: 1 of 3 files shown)


R1kaB3rN:
Great. Also, after working on zipapps, I am now considering to remove` umu_version.json` from the repo. Instead, we can simply add metadata such as the version to ` __init__.py`. If some of these steps can be automated by hatch, that would be great.

R1kaB3rN:
Closing as superseded by https://github.com/Open-Wine-Components/umu-launcher/pull/289

PR #175, opened 2024-09-05
feat: support running native Linux executables within SLR

R1kaB3rN:
Currently, we were running every executable through Proton (wine). However, Proton is a mechanism that is intended for running **Windows**  executables in predictable environment, not Linux ones. If the user or the client can determine that the game is a Linux executable and is confident that it can run reliably in the SLR, they can skip the overhead of wine. In the Steam client, users can enable this mechanism by forcing the Steam Linux Runtime (3.0) compatibility tool in their game's properties. As a result, the game will run inside sniper directly.

This feature is useful for games such as Katawa Shoujo which uses RenPy as its engine, hence do not need to be run through wine.

To run native Linux executables in the SLR, set `UMU_NO_PROTON=1`.

### DIFF
--- a/docs/umu.1.scd
+++ b/docs/umu.1.scd
@@ -168,6 +168,11 @@ _UMU_RUNTIME_UPDATE_
 
 	Set _0_ to disable updates.
 
+_UMU_NO_RUNTIME_
+	Optional. Runs the executable natively within the Steam Linux Runtime. Intended for native Linux games.
+
+	Set _1_ to run the executable natively within the SLR.
+
 # SEE ALSO
 
 _umu_(5), _winetricks_(1), _zenity_(1)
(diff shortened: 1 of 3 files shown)


loathingKernel:
I think the semantics are not clear as it is. `UMU_NO_RUNTIME` used to mean "without pressure-vessel", which is useful for debugging or *maybe* natively compiled Proton, if we ever want that. With these changes, if I understand correctly, it means "without proton but inside the runtime".  Probably the current env variable should be left as is, and introduce a new one, i.e. `UMU_NO_PROTON`, or even some kind of auto-detection based on `file` or similar utilities later on.

Disregarding the semantic issues, being able to run other non-proton steam compatibility tools with umu would be useful for tools such as Luxtorpeda. Luxtorpeda is a collection of open source game engines compiled for the SLR for older games. As such it should be run inside the SLR but not in Proton, overloading the meaning of `UMU_NO_RUNTIME` to be able to achieve that seems counter-intuitive.

Lampwick:
Does there need to be an option to select the runtime as in Steam? Sniper/Scout?

Issue #210, opened 2024-10-09
permission denied error


jagiella:
I install umu-launcher deb packages on ubuntu 24.04.

Running the launcher always ends up with an permission error:

```bash
$ WINEPREFIX=~/path/to/wineprefix/ GAMEID=umu-dauntless PROTONPATH=~/.local/share/Steam/compatibilitytools.d/UMU-Proton-9.0-3 umu-run ~/path/to/game.exe
steamrt is up to date
pressure-vessel-wrap[68721]: E: Der Kindprozess wurde mit Status 1 beendet: bwrap: setting up uid map: Permission denied
```

jagiella:
A workaround that solves that problem is tinkering with apparmor: https://github.com/DevToys-app/DevToys/issues/1198#issuecomment-2161540748

But this should not be a prerequist for the launcher to work, right?

GloriousEggroll:
this will always be a requirement on  ubuntu due to Ubuntu developers deciding to use Apparmor to restrict the creation of user namespaces: https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces

it may be possible to package an apparmor profile as part of the deb, something we need to look into

GloriousEggroll:
apparmor profile added https://github.com/Open-Wine-Components/umu-launcher/pull/218

gegarcia:
As @darix has commented on #218, please don't create an "unconfined" profile for bwrap. This completely bypasses the [unprivileged user namespace creation restriction](https://gitlab.com/apparmor/apparmor/-/wikis/unprivileged_userns_restriction) and makes the machine exploitable by attacks that use it.

A profile for scripts/programs that call bwrap should be created instead. Please [contact the AppArmor community](https://gitlab.com/apparmor/apparmor/-/issues) if you need any assistance.

AppArmor upstream has created a profile for bwrap that does not make your machine vulnerable. You can find the profile [here](https://gitlab.com/apparmor/apparmor/-/blob/master/profiles/apparmor/profiles/extras/bwrap-userns-restrict?ref_type=heads). It is not enabled by default on Ubuntu, but you can enable it by either copying the file from gitlab,  installing it on /etc/apparmor.d/, and load it using `sudo apparmor_parser /etc/apparmor.d/bwrap-userns-restrict`.

You can also use the version shipped by the apparmor-profile package, and create a symbolic link from the extra-profiles directory to /etc/apparmor.d/
[...]

PR #218, opened 2024-10-11
Add bwrap apparmor profile to debian packaging

CosmicFusion:


darix:
JFYI: i discussed this apparmor profile with the openSUSE apparmor maintainer. profiles like this were requested upstream multiple times and rejected.

in the mean time they added apparmor/profiles/extras/bwrap-userns-restrict which limits what bwrap started programs can do afterwards. 

so instead of copying a custom profile in, why not just enable that extra profile by packaging the needed symlink?

GloriousEggroll:
> JFYI: i discussed this apparmor profile with the openSUSE apparmor maintainer. profiles like this were requested upstream multiple times and rejected.
> 
> in the mean time they added apparmor/profiles/extras/bwrap-userns-restrict which limits what bwrap started programs can do afterwards.
> 
> so instead of copying a custom profile in, why not just enable that extra profile by packaging the needed symlink?

if it works it's more than welcome, feel free to make a PR. we simply were going based off results from this bug: https://github.com/Open-Wine-Components/umu-launcher/issues/210

I'm also not sure if that profile is included in ubuntu

darix:
the bwrap-userns-restrict profile might even be shipped in ubuntu but not turned on by default (as it is an extra profile)

I am not author of https://gitlab.com/apparmor/apparmor/-/blob/master/profiles/apparmor/profiles/extras/bwrap-userns-restrict?ref_type=heads 

but i can ask if they are fine with including the profile in umu-launcher. (gpl2 vs gpl3)

darix:
apparmor upstream suggested a better solution might be to apparmor profile the scripts that call bwrap in the umu scope and then you have full control about what the bwrap in the scope of umu should be able to do without affecting other tools like flatpak system wide. 

thoughts?

PR #229, opened 2024-10-14
fix: use $HOME/.local/share/umu and $XDG_CACHE_HOME for flatpak

### LINKED ISSUE #230: Games using Proton-GE (UMU) will not lauch in lutris flatpak
When trying to run the any game via the Proton-GE option in lutris  ie UMU launcher the game exits out with exit code 32000
the same issue doesnt occur with Wine GE 8.26
Lutris run time log - 

Started initial process 47 from gamemoderun /home/zerotwo/.var/app/net.lutris.Lutris/data/lutris/runtime/umu/umu_run.py /home/zerotwo/Games/ELDEN RING/Game/eldenring.exe
Start monitoring process.
[1mSetting up Unified Launcher for Windows Games on Linux...[0m
[1mUMU-Proton is up to date[0m
[1mDownloading latest steamrt sniper, please wait...[0m
[1mSteamLinuxRuntime_sniper.tar.xz: SHA256 is OK[0m
[1mVerifying integrity of sniper_platform_0.20241008.104210...[0m
[1msniper_platform_0.20241008.104210: mtree is OK[0m
pressure-vessel-wrap[57]: W: /dev/shm not shared between app instances (flatpak#4214). The Steam Overlay will not work.
steam-runtime-launch-client[57]: E: Unable to convert /app fd 16 into path: different file inside and outside sandbox
Monitored process exited.
Initial process has exited (return code: 32000)
All processes have quit
Exit with return code 32000

R1kaB3rN:
Centralizes the installation path of umu for both Flatpak and native applications in $HOME/.local/share/umu. This change will prevent cases like https://github.com/flathub/net.lutris.Lutris/issues/440 where the meaning of ~/.var/app/org.openwinecomponents.umu.umu-launcher/data/umu changes when the host's home directory is exposed.

Flatpaks will be required to change their manifest to allow read-write access to the host ~/.local/share/umu.

Closes https://github.com/Open-Wine-Components/umu-launcher/issues/230

### DIFF
--- a/umu/umu_consts.py
+++ b/umu/umu_consts.py
@@ -16,10 +16,25 @@ PROTON_VERBS = {
     "getnativepath",
 }
 
-# Installation path of the runtime files.
-# Flatpak will be detected as outlined by systemd
+XDG_DATA_HOME: Path = (
+    Path(os.environ["XDG_DATA_HOME"])
+    if os.environ.get("XDG_DATA_HOME")
+    else Path.home().joinpath(".local", "share")
+)
+
+# Installation path of the runtime files that respects the XDG Base Directory
+# Specification and Systemd container interface.
 # See https://systemd.io/CONTAINER_INTERFACE
-UMU_LOCAL: Path = Path.home().joinpath(".local", "share", "umu")
+# See https://specifications.freedesktop.org/basedir-spec/latest/index.html#basics
+# NOTE: For Flatpaks, the runtime will be installed in $HOST_XDG_DATA_HOME
+# then $XDG_DATA_HOME, and will be required to update their manifests by adding
+# the permission 'xdg-data/umu:create'.
+# See https://github.com/Open-Wine-Components/umu-launcher/pull/229#discussion_r1799289068
+UMU_LOCAL: Path = (
+    Path(os.environ.get("HOST_XDG_DATA_HOME", "XDG_DATA_HOME")).joinpath("umu")
+    if os.environ.get("container") == "flatpak"  # noqa: SIM112
+    else XDG_DATA_HOME.joinpath("umu")
+)
 
 # Temporary directory for downloaded resources moved from tmpfs
 UMU_CACHE: Path = (
--- a/umu/umu_consts.py
+++ b/umu/umu_consts.py
@@ -28,6 +28,10 @@ XDG_CACHE_HOME: Path = (
     else Path.home().joinpath(".cache")
 )
 
+_HOST_XDG_DATA_HOME: str = os.environ.get(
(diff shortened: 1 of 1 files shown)


umu/umu_consts.py
```diff
@@ -16,26 +16,14 @@
...
 # Flatpak will be detected as outlined by systemd
 # See https://systemd.io/CONTAINER_INTERFACE
-UMU_LOCAL: Path = (
-    Path.home().joinpath(
-        ".var", "app", "org.openwinecomponents.umu.umu-launcher", "data", "umu"
-    )
-    if os.environ.get("container") == "flatpak"  # noqa: SIM112
-    else Path.home().joinpath(".local", "share", "umu")
-)
+UMU_LOCAL: Path = Path.home().joinpath(".local", "share", "umu")
```
imLinguin: This should prefer `HOST_XDG_DATA_HOME` in container, and then fallback to the default.
Outside of the container it should also respect `XDG_DATA_HOME`  
Flatpak manifests then should add permission to `xdg-data/umu:create`

umu/umu_consts.py
```diff
@@ -16,10 +16,25 @@
...
+# Specification and Systemd container interface.
 # See https://systemd.io/CONTAINER_INTERFACE
-UMU_LOCAL: Path = Path.home().joinpath(".local", "share", "umu")
+# See https://specifications.freedesktop.org/basedir-spec/latest/index.html#basics
+# NOTE: For Flatpaks, the runtime will be installed in $HOST_XDG_DATA_HOME
+# then $XDG_DATA_HOME, and will be required to update their manifests by adding
+# the permission 'xdg-data/umu:create'.
+# See https://github.com/Open-Wine-Components/umu-launcher/pull/229#discussion_r1799289068
+UMU_LOCAL: Path = (
+    Path(os.environ.get("HOST_XDG_DATA_HOME", "XDG_DATA_HOME")).joinpath("umu")
```
imLinguin: This seems to be incorrect, the `"XDG_DATA_HOME"` will be used literally
ex.
```python
>>> env = {"some_key": "value"}
>>> env.get("HOST_XDG_DATA_HOME", "XDG_DATA_HOME")
'XDG_DATA_HOME'
```
`Path.home().joinpath(".local", "share")` seems like a reasonable default here. After all we are trying to access host's data directory here. If HOST variable is not specified, then we want the default `$HOME/.local/share/umu`

R1kaB3rN: Yeah, some environments (e.g., sway) may not set all of the XDG environments variables so we need to fall back to $HOME/.local/share/umu

strycore (approved):

PR #257, opened 2024-10-27
workflows: add workflow for snap

R1kaB3rN:
Related to https://github.com/Open-Wine-Components/umu-launcher/issues/299


vrement:
I was asking about the snap package in the other thread. Does this thread indicate that a ready snap package is in the works for normal users? 

I don't think many common people are able to build their packages themselves. And i for one can't be bothered. It's just too much.

PR #276, opened 2024-11-14
Revise workflows to automatically upload artifacts on new releases

### LINKED ISSUE #271: Feature request: provide package files directly, not as zip files
I noticed that for the last couple releases the various distribution packages have come zipped up instead of the “naked” packages being attached to the release.

I’m assuming there is a reason for it, but running `dnf install <url>` is way more convenient than downloading an archive, unpacking it, and only _then_ installing it :)

loathingKernel:
This PR revises the GH workflows to upload build artifacts to releases automatically. It should fix #271 among other things. Once a release is created and published the `release` workflow should build the artifacts and upload them automatically.

Changes:
* Canonicalize artifact names for different distributions
   - On RPM packages, the version information in the package name was truncated to remove the date and short SHA information. This change would be a better fit for the spec file but I didn't want to touch it yet.
   - On DEB packages, the `_ubuntu-noble` or `_debian-12` suffixes are used to differentiate between them. @CosmicFusion is this accepted by Debian packaging tools?
   - The zipapp name was changed to contain version information and compressed with `gzip`, i.e `umu-launcher-1.1.4-zipapp.tar.gz`. Third-party scripts will need to be adjusted.
* The build workflows for RPM and DEB packages have been adjusted to also build development snapshots. This requires copious amounts of `sed` and I would be more than happy to revert these changes and only build tagged releases. The RPM spec was also adjusted for the same reason to checkout the `manual_commit` if the `tag` doesn't exist.
* Added a separate development workflow to build packages and upload them only to the action's artifacts, The individual packaging workflows are not executed on push or PRs any more, but through this development `build` workflow.

The workflow for RPM packages fails currently because cloning the repo inside the specfile doesn't find the PR merge commit. If we want to handle that for PRs it will take some finessing, but I want to be sure it is needed before diving into it.

image

### DIFF
--- a/.github/workflows/build-umu-debian-12.yml
+++ b/.github/workflows/build-umu-debian-12.yml
@@ -1,11 +1,11 @@
 name: UMU Deb Build - Debian 12
 on:
-  push:
-    branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
-  release:
-    types: [published]
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+
 
 jobs:
   build:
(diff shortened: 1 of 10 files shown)


CosmicFusion:
>    - On DEB packages, the `_ubuntu-noble` or `_debian-12` suffixes are used to differentiate between them. @CosmicFusion is this accepted by Debian packaging tools?

My original PR didn't differentiate between distros, nevertheless apt will except a package aslong as it ends in 
```
.deb
```
But the suffixes course will not mean anything to apt.

PR #305, opened 2024-12-09
Do not force the runtime for winetricks if `UMU_NO_RUNTIME` is set

loathingKernel:


merge-enjoyer:
@R1kaB3rN, it is time to merge the code. Ty.

R1kaB3rN:
UMU_NO_RUNTIME is an internal environment variable and winetricks in the context of umu is a Proton property, which is itself a property of the Steam Linux Runtime. The pull request implies that UMU_NO_RUNTIME is being used as a feature, and winetricks is being used outside its property. From that logic, running it outside its property is equivalently problematic as running Proton outsides its property, risking more bug reports. Closing.

loathingKernel:
We have been using umu with `UMU_NO_RUNTIME` with proton-cachyos for sometime now to make it easily available in lutris. Running proton-cachyos inside the steam runtime when it is natively compiled is not possible. This allows to us to run winetricks with proton-cachyos without issues in that case. In this use-case Proton is not a property of Steam Linux Runtime and `winetricks` itself is being distributed as part of `umu-protonfixes`, making the SLR irrelevant in the process itself.

Furthermore, UMU_NO_RUNTIME has been used with umu's flatpak, is it acceptable in that case to run winetricks using the runtime?

PR #306, opened 2024-12-09
Do not check if winetricks settings verbs are installed

loathingKernel:
Umu would complain and refuse to go back and forth between settings, such as `sound=alsa` and `sound=pulse`, as they were written in `winetricks.log`. This also becomes a problem when installing verbs that set a specific windows version, such as `vcrun2019` which sets `winxp`, as it is impossible to reset the version to `win10` after the first time. Leaving it to `winxp` is not an option as it can cause games to fail to launch with dx12.

Initially I wanted to do this dynamically and check against the active `winetricks` version, but since the check is before the environment is set up, it would require more changes. As a result, the winetricks verbs have been dumped as a static set in `umu_consts.py` to check against. Dumping script has been included for future use, and could serve as part of the dynamic check.


R1kaB3rN:
To clarify, the decision to complain was implemented to prevent reentering the container, Proton’s prefix setup, and installing the same stuff twice. However, since you’re saying this is problematic and that there are cases to go back and worth, perhaps we should get rid of it all together and give clients that freedom. Or alternatively, add an environment variable to force the install

loathingKernel:
I think complaining is good for `dlls` and such, but not useful for `settings`. Winetricks itself will complain for dlls and require to be forced to install them again. Settings on the other hand are exempt from such checks. In that regard, `winetricks` is better than umu in identifying "conflicting" dlls, for example it will complain if the user tries to install `vcrun2022` but `vcrun2019` is already installed.

Multiple concurrent usages of `winetricks` is indeed an issue, but this could possibly be addressed by a mechanism that locks the prefix. Possibly as part of a mechanism to allow automatic verb fallback (`waitforexitandrun` -> `runinprefix` if something that is not winetricks runs in the prefix already)

Issue #134, opened 2024-07-03
Option to disable auto updates on startup


AdamantGarth:
I'd like to be able to disable steamrt/proton updates on umu startup (using an environment variable, for example), and trigger them manually from time to time.

R1kaB3rN:
I believe we want users to always use the latest runtime and Proton, and if we add an option it'll need to be sometime after the next release.

Though, would you still wish for this feature if the updates for those builds were faster? Because I would understand the annoyance for waiting for the updates for those builds to complete before running the game.

AdamantGarth:
> I believe we want users to always use the latest runtime and Proton

As a user I do too, and I'm very happy that umu can update these things itself. But sometimes I have to use metered connections that are both too slow and too expensive to be downloading hundreds of megabytes of runtimes on. I doubt that there's a way to reduce the size of these runtimes to be both fast and cheap to download on such connections, so an option to disable the auto updates would be much appreciated.

db47h:
Another use case for this is when your broadband connection goes down: umu just crashes on a failing DNS query (my local net was still up but no external name resolution would work). Although for that specific case a more graceful handling of network errors would be nice, an option to disable updates would probably be simpler to implement.

gabriele2000:
Guys, just hear me out: DELTA UPDATES

R1kaB3rN:
Graceful handling of network errors will be explored in the future. Same with delta updates once we start [customizing the runtime](https://github.com/Open-Wine-Components/umu-launcher/issues/4), as the delta file would need to be created at build time.

Issue #133, opened 2024-07-03
This is more like suggestion instead of issue - problem with GE-Proton solution is similar to umu-proton


rairay91:
When i run a game with native umu-proton the launcher automatically check for update for latest umu-proton release, if there is a new release umu download it and create a link umu-latest/umu-proton whatever to that version and delete old. Now my issue is that for working with ge-proton which is more uptodate and have more patches/fixes than umu-proton, i have to manage it manually. Umu download latest version automatically at first boot but keep older version and do not create a link to the latest ge-proton just like it does to latest umu-proton.
So at the end umu should also download and create GE-Proton link to the latest downloaded GE-Proton in the same directory as it does to the latest umu-proton and should delete older version unless i assigned umu to use a specific version of proton.
Why downloading latest version of ge-proton then creating a link to that in the same directory and deleting previous version is important?
Because it eliminates the overhead of managing this manually.
And to be honest keeping older version of ge-proton is not necessary because i haven't ever noticed if we need to keep older ge-proton. Even if there is a game that need old version of ge-proton, we can do it manually.

loathingKernel:
The basic idea is that UMU-Proton can be a stable version as close as possible to Valve's stable Proton that includes the minimum required modifications to work with umu, and as such it serves as the default version to download. On the other hand GE-Proton has a lot more patches, a lot more customization and it is more likely for a broken update to occur.

>  Because it eliminates the overhead of managing this manually.

This is all well and good if nothing breaks, but if something does break, and a user wants to test different versions, then umu should not get in their way of testing the older versions. If by default umu would update or even delete older GE-Proton versions it would only add to the confusion and the frustration of the user who is trying to identify and solve an issue.

There are other very good tools that can do what you want to do such as protonup-qt, the GUI launcher of your preference, or even your own custom update script. There is very little reason to burden umu with that functionality.

rairay91:
Your answer seems helpfull. But i barely noticed any game that exclusive requires specific proton version. But that is okay.
However, you can do one thing. Whenever there is a new version of proton-ge downloaded by umu, umu automatically assign a link to the latest version of ge-proton with specific name like "GE-Proton", similar to how it does with umu-proton. In this situation i can specify to all or most of my games to use that proton i.e. "GE-Proton".
[...]

PR #308, opened 2024-12-10
feat: add initial support for delta updates

R1kaB3rN:
Related to https://github.com/Open-Wine-Components/umu-launcher/issues/133#issuecomment-2209551963 and https://github.com/Open-Wine-Components/umu-launcher/issues/134.

Adds optional support for delta updates for currently supported compatibility tools, resulting in significantly faster updates to the latest releases. This also puts umu-launcher up to parity with the Steam client in terms of update distribution for its tools.

For clients, this can be enabled by setting either `PROTONPATH=GE-Latest` or `PROTONPATH=UMU-Latest` where each of those values represent the latest GE-Proton and UMU-Proton respectively. In this system, those builds will be placed within the private directory $XDG_DATA_HOME/umu/compatibilitytools, and like Steam's Proton Experimental, only 1 build will be maintained. For this system to work, it will be expected for clients or users to not modify the contents of those builds.

Delta updates to the Steam Linux Runtime will be supported later after we start building it, and will become the default. Additionally, this feature is currently not supported for zipapps.

For this feature, `cbor2`, `pyzstd` and `xxhash` will be a new optional dependencies. Additionally, `cargo` will be a new build dependency.

TODO:
- Refactor code to pull patch file from Github or OWC controlled server


R1kaB3rN:
Will resolve the Debian build failure in a bit and write the basic tests later. For testing, this suffices for now.

R1kaB3rN:
@beh-10257 I'm interested to know if the build succeeds on NixOS. Do you mind testing how this builds on NixOS? I'd appreciate it.

Also, if you're interested in testing this, I'll attach the patch file intended to upgrade GE-Proton9-20 below in a bit. To test, first install `cbor2` then install/copy GE-Proton9-20 to `$HOME/.local/share/umu/compatibilitytools` then rename it to **GE-Latest**. If you have it installed already and are copying it over, be sure to delete all the `__pycache__`.

Afterwards, assuming you have the patch in your home directory, run:
`PROTONPATH=GE-Latest GAMEID=umu-0 UMU_DELTA=$HOME/delta.cbor umu-run ""`

R1kaB3rN:
https://drive.proton.me/urls/18EM1D76FG#VlasLi0bh7Qa

bohanubis:
ok will do will build this branch thanks also can I send a patch that adds cbor2 as optional dependency
(to make my life easier now)

R1kaB3rN:
> ok will do will build this branch thanks also can I send a patch that adds cbor2 as optional dependency (to make my life easier now)

It's fine right now, because cbor2 is already an optional dependency.

See https://github.com/Open-Wine-Components/umu-launcher/blob/dfd85aa67faa6a2898b8b5b7e9c24def3e2305c4/pyproject.toml#L45

R1kaB3rN:
In other words, it will not be installed by default for the system package umu-launcher. Though, we'll probably need to make it clearer in the Nix flake. Because cbor2 will be needed to activate delta updates

R1kaB3rN:
In the Flake, the latest stable version of Rust will need to be added as a build dependency too. Perhaps that's called `pkgs.rustc`?

bohanubis:
I mean I added cbor2 as an optional dependency I mean I can turn it off by default but why not
heres a patch for cbor2
wait I can't upload .patch files ???

bohanubis:
> In the Flake, the latest stable version of Rust will need to be added as a build dependency too. Perhaps that's called `pkgs.rustc`?

as an optional or a main dependency

bohanubis:
[0001-add-cbor2-as-an-optional-dependency-for-nixos.txt](https://github.com/user-attachments/files/18241810/0001-add-cbor2-as-an-optional-dependency-for-nixos.txt)

just rename .txt to .patch

bohanubis:
merge that patch into the branch
I will test the delta updates in around 30 minutes
I'm downloading the cbor2 file right now
also
image
I have this can't I copy ge-proton-9-20 and see cbor actually do its magic
or will it have to download from scratch to do its thing
also what is ??
`__pycache__`
aka where is the location

R1kaB3rN:
> > In the Flake, the latest stable version of Rust will need to be added as a build dependency too. Perhaps that's called `pkgs.rustc`?
> 
> as an optional or a main dependency

I meant Rust as a build dependency in depsBuildBuild.

https://github.com/Open-Wine-Components/umu-launcher/blob/6189d0d9fd062e89a375db20aeae1d1c009e9833/packaging/nix/umu-launcher.nix#L8

bohanubis:
feel free to add rustc (on a sidenote) cargo is not included so if its being used add it as well

R1kaB3rN:
> merge that patch into the branch I will test the delta updates in around 30 minutes I'm downloading the cbor2 file right now also image I have this can't I copy ge-proton-9-20 and see cbor actually do its magic or will it have to download from scratch to do its thing also what is ?? `__pycache__` aka where is the location

The python interpreter in the runtime creates a `__pycache__` folder that stores its compiled bytecode. It does this so it'll run faster for future launches, and it's also part of the reason why launching a new downloaded build can be a bit slow. In GE-Proton9-20, you should see folders like `__pycache__`.

R1kaB3rN:
If you have the fd program, within GE-Proton9-20, you can run the command ` fd '__pycache__' . -X rm -r` to get rid of it.

bohanubis:
image
I mean there is nothing

R1kaB3rN:
> feel free to add rustc (on a sidenote) cargo is not included so if its being used add it as well

Yeah, cargo is needed. Not just rustc, and I can add that.

bohanubis:
after searching for pycache in my entire home steam is the only program that actually creates those folders
no mention of umu so we are good to go I guess

R1kaB3rN:
> image I mean there is nothing

Strange, unless you're using bwrap that binds that path as read-only? If that's the case then the `__pycache__` won't persist and you'll be regenerating it each launch

R1kaB3rN:
> after searching for pycache in my entire home steam is the only program that actually creates those folders no mention of umu so we are good to go I guess

Alright. Yeah, so if there's not cache go ahead and copy GE-Proton9-20 to $HOME/.local/share/umu/compatibilitytools then rename it to GE-Latest.

bohanubis:
> > image I mean there is nothing
> 
> Strange, unless you're using bwrap that binds that path as read-only? If that's the case then the `__pycache__` won't persist and you'll be regenerating it each launch

what is read only ??
the proton directory in the bwrap of umu ??

bohanubis:
> > after searching for pycache in my entire home steam is the only program that actually creates those folders no mention of umu so we are good to go I guess
> 
> Alright. Yeah, so if there's not cache go ahead and copy GE-Proton9-20 to $HOME/.local/share/umu/compatibilitytools then rename it to GE-Latest.

ok will do (I'm waiting for you to merge that patch tbh)
and also redownloading ge proton 9-21 to see if pycache shows up

R1kaB3rN:
> > > image I mean there is nothing
> > 
> > 
> > Strange, unless you're using bwrap that binds that path as read-only? If that's the case then the `__pycache__` won't persist and you'll be regenerating it each launch
> 
> what is read only ?? the proton directory in the bwrap of umu ??

Yeah, that's what I meant. Since I know you use bwrap, unless you're not doing anything special with the proton.

R1kaB3rN:
Should be good now. I applied your patch and added cargo as a build dependency.

bohanubis:
PROTONPATH=GE-Latest GAMEID=umu-0 UMU_DELTA=$HOME/delta.cbor umu-run ""

do I need the umu_delta bit since I have ge proton 20

bohanubis:
update it doesn't build since cargo access the internet
and thats impure

is cargo only used for bzip2

R1kaB3rN:
Cargo is used to fetch all of the necessary dependencies. See Cargo.toml for details.

Are you not able to do that in NixOS?

R1kaB3rN:
> PROTONPATH=GE-Latest GAMEID=umu-0 UMU_DELTA=$HOME/delta.cbor umu-run ""
> 
> do I need the umu_delta bit since I have ge proton 20

Currently, in this PR, you need to set UMU_DELTA. Once this is merged, that won't be a thing, and umu-launcher will fetch the patch from our server.

bohanubis:
I mean there are a bunch of rust applications in nix so it will work out in the end
I just don't know how on the top of my head
probably I'm gonna fix this tomorrow

R1kaB3rN:
We're not depending and using any Rust apps/packages. Instead we're using Rust libraries.

bohanubis:
I have in my mind 2 solutions (use whatever the proper way of handling cargo.toml in nix)
or just provide said libraries from nixpkgs
I will ask more experienced nix people for help and be back
probably I will fix this tomorrow
sorry

R1kaB3rN:
cc @LovingMelody for insight.

So for this feature, Rust will be required to build a module. In the Nix build, it shouldn't require anything special besides adding Cargo, right?

bohanubis:
> cc @LovingMelody for insight.
> 
> So for this feature, Rust will be required to build a module. In the Nix build, it shouldn't require anything special besides adding Cargo, right?

so rust will build a python module that umu will be dependant upon
aka umu is not directly dependant on those rust libraries ??

LovingMelody:
The cargo dependencies would need to be pinned as well. I don't see a Cargo lock file included in this pr (on mobile so might of missed it). 

Linking the nix docs for rust for reference 

https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md

R1kaB3rN:
> > cc @LovingMelody for insight.
> > So for this feature, Rust will be required to build a module. In the Nix build, it shouldn't require anything special besides adding Cargo, right?
> 
> so rust will build a python module that umu will be dependant upon aka umu is not directly dependant on those rust libraries ??

At runtime, umu becomes directly dependent on the Rust libraries if cbor2 is installed, if that makes sense. Right now, the Rust modules are being built and included by default. This way, the user just needs to install cbor2 to use this feature.

R1kaB3rN:
> The cargo dependencies would need to be pinned as well. I don't see a Cargo lock file included in this pr (on mobile so might of missed it).
> 
> Linking the nix docs for rust for reference
> 
> https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md

Thanks. So from what I gleaned from the documentation, using a fetcher will be required and it looks like adding the Cargo.lock file to the Flake will be necessary.

https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md#vendoring-of-dependencies-vendoring-of-dependencies

bohanubis:
image
morning

just make it something like this

bohanubis:
image
also I think its fair to say this is not working
image

also I tried download 9-21 using this fork and __pycache__ is still not generating

can you make it so it throws an error if it can't generate __pycache__

bohanubis:
also I have an idea of a check for nixos that works on any distro
steps:
1_ use a distro that has nix available arch whatever
2_ git clone this repo
3_ cd into packaging/nix directory
4_ run this command
`nix flake update && nix --extra-experimental-features 'nix-command flakes' build ./#umu -L`
if its a success than great if not than I guess tag me
also `nix flake update` can fail due to network issues so if we can replace that part with try 5 times if it doesn't work move on to
`nix --extra-experimental-features 'nix-command flakes' build ./#umu -L`
that would be ideal

R1kaB3rN:
Pycache isn’t a big deal in the context of this PR, and I’ll leave that to you to explore on why it’s not being created.

R1kaB3rN:
> image morning
> 
> just make it something like this

Thanks and I’m going to assume that it builds fine after I make that change

LovingMelody:
> image
> 
> also I think its fair to say this is not working
> 
> image
> 
> 
> 
> also I tried download 9-21 using this fork and __pycache__ is still not generating
> 
> 
> 
> can you make it so it throws an error if it can't generate __pycache__

Not sure why you are so fixated on pycache being generated should be a no issue. 

https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/interpreters/python/setup-hook.sh

R1kaB3rN:
Refactored the code to handle the new patch format as I saw decompression to be a big bottleneck. Applying the patch will be more efficient now as zstd is being used as the engine.  

To that end, python-pyzstd and python-xxhash will be new optional dependencies. The former is pretty new so it will be vendored and isn't available on NixOS. Later, I can try to see if I can use the Rust library equivalents. That way only libzstd.so and libxxhash.so will be required

R1kaB3rN:
Link to new patch (GE-Proton9-21 -> GE-Proton9-22):

- https://drive.proton.me/urls/GXMAWNPN7C#YlmhptkSB7xM

Edit: Updated link

packaging/nix/umu-launcher.nix
```diff
@@ -13,14 +13,21 @@ python3Packages.buildPythonPackage {
...
   propagatedBuildInputs = [
     pyth1
     pkgs.bubblewrap
     pkgs.python3Packages.xlib
     pkgs.python3Packages.filelock
     pkgs.python3Packages.urllib3
-  ] ++ lib.optional truststore pkgs.python3Packages.truststore;
+  ] ++ lib.optional truststore pkgs.python3Packages.truststore
+    ++ lib.optional cbor2 pkgs.python3Packages.cbor2
+    ++ lib.optional xxhash pkgs.python3Packages.xxhash;
```
LovingMelody: `xxhash` has been added as optional but has not been added to the arguments

LovingMelody (changes requested):

packaging/nix/umu-launcher.nix
```diff
@@ -1,4 +1,4 @@
-{lib, pyth1 ,python3Packages , umu-launcher, pkgs,version, truststore ? true, ...}:
+{lib, pyth1 ,python3Packages , umu-launcher, pkgs,version, truststore ? true, cbor2 ? true, xxhash ? true, rustPlatform, ...}:
```
LovingMelody: nitpick: cbor2 & xxhash might be better represented as `deltaUpdates ? true`

LovingMelody (approved):
Didn't test binary but confirmed this is building.

packaging/nix/umu-launcher.nix
R1kaB3rN: Added `deltaUpdates ? true`

R1kaB3rN:
I think zstd may also need to be included in the Flake. pyzstd depends on libzstd.so, and we don't build it. Instead, in the Makefile, we just download the module and depend on the system's libzstd.so.

R1kaB3rN:
Not going to commit Cargo.lock because it's causing CI failures for Ubuntu and Debian, so the Nix flake will need to be updated to not use it. The lock file version format seems to move with the Rust compiler version, and the latest format (v4) just breaks the builds for those distros. Manually editing the lock file to target v3 is an option, but I don't know if that's sane to do.

LovingMelody:
Seems odd that it is breaking the CI as it is standard to include lock files when creating binaries.
https://github.com/rust-lang/cargo/issues/315

Edit: new blog post https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

R1kaB3rN:
Read the articles. Yeah, definitely agree that it’d be best to commit Cargo.lock. I’ll do that.

R1kaB3rN:
Also, another option is to not use Rust. Instead, we could just use python3-cryptography for the crypto stuff. While lower quality, it’s practical and requires less. After all, we’re just a tool that enables playing video games

PR #309, opened 2024-12-14
docs(nix): pkgs.system -> pkgs.stdenv.hostPlatform.system

MattSturgeon:
General improvements & expansions to the nix installation instructions. I've gone over the nix installation instructions and added a few minor cleanups and expansions, including:

`pkgs.system` is not generally recommended. It's usually better to explicitly using the "host platform" system from `pkgs.stdenv`.

This is required to support cross-compile, for example.

For example, we can use nix's preferred flakeref syntax instead of a `git+` URI.

Also, `pkgs.system` is soft-deprecated: for example, using `stdenv.hostPlatform` will allow configurations to be remote/cross compiled. Therefore I've switched the example to using this.

I tried to make things a little more readable using a `let-in` block.

I added a note that `inputs` must be manually passed from your flake to your nixos configuration eval.

I updated the existing note to use [github-flavoured markdown admonitions](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts).

Note: I've kept this as several very atomic commits so that it may be easier to review and/or drop any changes that are not wanted. Happy to squash some/all or leave as-is; please advise :upside_down_face: 

cc @beh-10257

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ and in your `configuration.nix`
 let
   inherit (pkgs.stdenv.hostPlatform) system;
   umu = inputs.umu.packages.${system}.umu.override {
-    version = "${inputs.umu.shortRev}";
+    version = inputs.umu.shortRev;
     truststore = true;
   };
 in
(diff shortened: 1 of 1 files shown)


bohanubis:
pretty much thats all thanks a lot for the pr
I didn't know about pkgs.system being deprecated so my bad yeah thanks again

PR #311, opened 2024-12-14
Bump version to 1.2.0

R1kaB3rN:


R1kaB3rN:
Draft for now. After inspecting the GitHub artifact, it was discovered the RPM package is missing the vendored urllib3 and the Rust module. Will dig into later

R1kaB3rN:
The vendored urllib3 and Rust module issue was fixed. The only thing holding back the release are the .deb packages. See https://github.com/Open-Wine-Components/umu-launcher/pull/330

R1kaB3rN:
The issue with the .deb packages was fixed.

PR #313, opened 2024-12-23
fix: provide bwrap-userns-restrict for debian

### LINKED ISSUE #280: Debian package doesn't provide bwrap-userns-restrict when built
After building umu-launcher 1.1.4 on Debian Unstable and installing the resulting packages the following error appears:

```
Setting up python3-umu-launcher (1.1.4-1) ...
Job for apparmor.service failed because the control process exited with error code.
See "systemctl status apparmor.service" and "journalctl -xeu apparmor.service" for details.
```

Checking the journal shows:

```
There was an error while loading profiles from /etc/apparmor.d
```

Checking `/etc/apparmor.d` shows that there's a broken symlink:

```
$ ll "$(readlink /etc/apparmor.d/bwrap-userns-restrict-umu)"
ls: cannot access '/usr/share/apparmor/extra-profiles/bwrap-userns-restrict': No such file or directory
```

R1kaB3rN:
Fixes https://github.com/Open-Wine-Components/umu-launcher/issues/280 as suggested from upstream maintainer at https://github.com/Open-Wine-Components/umu-launcher/issues/210#issuecomment-2411412865.


ChthonVII:
Apparmor in Debian 12 (3.1.7-1+b3) chokes on that file and fails to start with "AppArmor parser error for..."

PR #322, opened 2025-01-07
Nix fixes

fufexan:
- format all nix files using alejandra. I can add it as the default formatter as well. If you prefer a different formatter, that can also be done.
- structure packages inside overlay for easier management and overrideability (especially considering the recent hatch breakage)
- update lockfile (needed for rust 1.83)
- temporary fix for hatch build failure (can be reverted once https://github.com/NixOS/nixpkgs/issues/366359 is solved)

### DIFF
--- a/packaging/nix/combine.nix
+++ b/packaging/nix/combine.nix
@@ -1,8 +1,19 @@
-{ env, package, symlinkJoin,version,truststore,cbor2 }:
+{
+  env,
+  package,
+  symlinkJoin,
+  version,
+  truststore,
+  cbor2,
+}:
 symlinkJoin {
   name = "umu-run-bwrap";
   paths = [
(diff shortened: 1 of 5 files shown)


R1kaB3rN:
Looks good to me. Also, I created a [package request for pyzstd upstream](https://github.com/NixOS/nixpkgs/issues/371859). Once that's merged we can add another patch to not vendor pyzstd.

cc @beh-10257 just so that you're aware.

R1kaB3rN:
> format all nix files using alejandra. I can add it as the default formatter as well. If you prefer a different formatter, that can also be done.

Sure, that's fine. That way, there'll be a standard format which I think is good.

bohanubis:
nice thanks

PR #325, opened 2025-01-08
ci: fix mypy return Any

R1kaB3rN:
Related to https://github.com/Open-Wine-Components/umu-launcher/actions/runs/12677716535/job/35333669712

Trivial and not serious. Just cannot reproduce locally within the venv.

### DIFF
--- a/umu/umu_proton.py
+++ b/umu/umu_proton.py
@@ -138,7 +138,7 @@ def _fetch_patch(session_pools: SessionPools) -> bytes:
     if resp.status != HTTPStatus.OK:
         return b""
 
-    return resp.data
+    return resp.data  # type: ignore
 
 
 def _fetch_releases(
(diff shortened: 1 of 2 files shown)


R1kaB3rN:
Strange. Cannot reproduce using mypy 1.14.1 in a virtual environment. 

> > $ mypy --version
mypy 1.14.1 (compiled: yes)
> $ which mypy
/home/foo/umu-launcher/.venv/bin/mypy
> $ python --version
Python 3.10.16
> $ which python
/home/foo/umu-launcher/.venv/bin/python
>$ mypy umu/
Success: no issues found in 11 source files

R1kaB3rN:
There's 2k+ issues in the mypy issue tracker, and [querying with the relevant error string currently returns 34 issues related to this](https://github.com/python/mypy/issues?q=is%3Aissue+Returning+Any+from+function+declared+to+return). Ignoring, and won't fix.

PR #327, opened 2025-01-10
fix: guide for nixos install

### LINKED ISSUE #324: nix: rustc versioning, unable to build.
Getting this following the README instructions.
```
  > :: Building delta dependencies
       > cargo build -r
       > error: rustc 1.82.0 is not supported by the following package:
       >   umu@0.1.0 requires rustc 1.83
       >
       > make: *** [Makefile.in:176: builddir/.build-umu-delta] Error 101
```

I also have a system wide rust install via rustup and rustc is at 1.83
```
❯ : rustc --version
rustc 1.83.0 (90b35a623 2024-11-26)
```

kumawatdarshan:
closes #324 

NixOs stable(nixpkgs 24.11) rustc version is at 1.82 which was causing the issue. 
Tho, is 1.83 really needed? Rust MSRV should be lower for easier packaging upstream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,10 +147,7 @@ If there is any problem with the flake feel free to open a bug report and tag an
 If you want to add umu-launcher as a flake add this to your inputs in `flake.nix`
 ```nix
   inputs = {
-    umu = {
-      url = "github:Open-Wine-Components/umu-launcher?dir=packaging/nix";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
+    umu.url = "github:Open-Wine-Components/umu-launcher?dir=packaging/nix";
   }
 ```
 and in your `configuration.nix`


R1kaB3rN:
No, we don’t. Thanks for pointing that out. Yes, the MSRV is able to be downgraded as far as 1.81, which will allow Fedora and Nobara to use its system rustc.

R1kaB3rN:
Fixed the MSRV to something reasonable MSRV (1.65) for now at 414a1a8e630cd190a29120e511b1d7efd73328bd. Again, thanks for pointing that out.

cc @beh-10257 for awareness of this change.

kumawatdarshan:
I think this PR should be reverted as msrv has been downgraded. 
Since afaik `inputs.nixpkgs.follows = "nixpkgs";` is for package coherence, actual issue was just msrv.

PR #328, opened 2025-01-10
build: downgrade MSRV

R1kaB3rN:
Related to https://github.com/Open-Wine-Components/umu-launcher/pull/327

Enables NixOS users on stable (24.11) to build with its system rustc.


R1kaB3rN:
cc @beh-10257 

This change affects NixOS. The Cargo.lock file has been downgrade to v3 and generated with rustc 1.65. I don't expect this to be a problem though.

PR #332, opened 2025-01-13
build: fix missing urllib3 in RPM

R1kaB3rN:
The urllib3 vendored dependency was discovered to be missing in the RPM. This was because the RPM spec was pinned to 1.1.4, and was missing pip as a build dep. This fixes that and also adds the Rust module, which was fixed along the way.


R1kaB3rN:
Merging for now. There's a lot of room for improvement though, which can be addressed later.

PR #363, opened 2025-02-12
make: add options to allow disabling of builtin pyzstd and urllib in favor of native system versions

GloriousEggroll:
make: add options to allow disabling of builtin pyzstd and urllib in favor of native system versions

examples:
```
# Default behavior (build included components)
make

# Use native pyzstd but include urllib3
make USE_NATIVE_PYZSTD=xtrue

# Use native urllib3 but include pyzstd
make USE_NATIVE_URLLIB=xtrue

# Use both native components
make USE_NATIVE_PYZSTD=xtrue USE_NATIVE_URLLIB=xtrue
```

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -19,6 +19,11 @@ DESTDIR ?=
 USERINSTALL ?= xfalse
 FLATPAK ?= xfalse
 
+# For vendors that prefer to use native pyzstd and urllib3.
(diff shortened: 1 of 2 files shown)


R1kaB3rN:
[0001-deb-fix-build-by-using-rustup.txt](https://github.com/user-attachments/files/18762543/0001-deb-fix-build-by-using-rustup.txt)

PR #364, opened 2025-02-12
Improve nix section of the README

MattSturgeon:
General cleanup and improvements to the nix section of the README:

- Mention that umu-launcher is available in nixpgks (since 25.05, currently unstable)
- General cleanup
- Improved examples
  - Streamline examples
  - Added home-manager example
  - Fold examples within `<details>` blocks
  - Format examples using alejandra
- Re-structure the nix package overrides section
  - Document downstream package args
  - Re-structure documentation for our package args
  - Briefly mention `version` package arg

cc @beh-10257 @LovingMelody @R1kaB3rN

PR #365, opened 2025-02-12
Add an overlays example to the nix section of the README

MattSturgeon:
Using overlays can provide a simpler interface, as there's no need to specify `system` and all packages end up in `pkgs`.

This adds two similar examples, alongside the existing example that doesn't use overlays.

I didn't include this in #364 as it felt a bit beyond its scope.

**[Rendered markdown](https://github.com/MattSturgeon/umu-launcher/blob/docs/nix/README.md#nix)**

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,80 @@ in {
 
 </details>
 
+<details><summary><strong>Example flake</strong> (using overlays)</summary>
+
+If you pass your flake inputs to `specialArgs`, you can apply the overlay within your `configuration.nix`:
+
+```nix
+# flake.nix
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    umu.url = "github:Open-Wine-Components/umu-launcher?dir=packaging/nix";
+    umu.inputs.nixpkgs.follows = "nixpkgs";
+  };
+  outputs = {nixpkgs, ...} @ inputs: {
+    nixosConfigurations.desktop = nixpkgs.lib.nixosSystem {
+      # Pass `inputs` to the nixos configuration,
+      # allowing it to be used as a module arg
+      specialArgs = {inherit inputs;};
+      modules = [./configuration.nix];
+    };
+  };
+}
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -247,6 +247,80 @@ in {
...
+
+If you pass your flake inputs to `specialArgs`, you can apply the overlay within your `configuration.nix`:
+
+```nix
+# flake.nix
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    umu.url = "github:Open-Wine-Components/umu-launcher?dir=packaging/nix";
+    umu.inputs.nixpkgs.follows = "nixpkgs";
```
LovingMelody, line 260: Think we should leave a note here about stable channels missing packages. They won't be in a stable release for 3 months and future dependencies may also be missing.

MattSturgeon: Isn't that already covered by [L167](https://github.com/Open-Wine-Components/umu-launcher/blob/8bf01dd9ba5dd180a963776549acdc20265560e3/README.md?plain=1#L167) and [L195](https://github.com/Open-Wine-Components/umu-launcher/blob/8bf01dd9ba5dd180a963776549acdc20265560e3/README.md?plain=1#L195)?

Also, the `follows` is less relevant when using overlays, as umu's nixpkgs input will only be used for `nixpkgs.lib` in that scenario.

R1kaB3rN:
Question: In the README.md, why umu-launcer? https://github.com/Open-Wine-Components/umu-launcher/blob/aa33f7840c3af64a6fad5438455992cb2c17f3d6/README.md?plain=1#L167

R1kaB3rN:
If it's a typo, I don't mind correcting that here and would not sweat it being out of scope.

MattSturgeon:
> If it's a typo, I don't mind correcting that here and would not sweat it being out of scope.

Opened #369 to keep it separate, in case you end up squash-merging this or something.

R1kaB3rN (approved):

PR #374, opened 2025-02-14
Support overriding 1.2.0+ nix package

MattSturgeon:
Prepare for the upcoming version bump in nixpkgs, by making the unnecessary or conflicting overrides conditional on the version of the package being overridden.

This is compatible with nixpkgs inputs both with and without https://github.com/NixOS/nixpkgs/pull/381975, because we can check the base-package's version and choose an override impl based on that.

If the base package is older than 1.2.0, we use the old/current impl. If the package is 1.2.0 or newer, we can drop those overrides and only use the newer simpler impl.

I've split the old & new impl into separate files to simplify the diff and make the eventual removal easier.

### Testing

To test the "old" impl, you can simply clone this PR and run `nix build`.

To test the "new" impl, against https://github.com/NixOS/nixpkgs/pull/381975, you can override the `nixpkgs` input and then build:

```sh
nix flake lock --override-input nixpkgs 'github:nixos/nixpkgs?ref=pull/381975/merge'

nix build
```

> [!TIP]
> Use `ref=pull/<number>/merge` to test a PR merged with the latest nixpkgs master
> or use `ref=pull/<number>/head` to test the PR itself.

cc @beh-10257 @LovingMelody @diniamo @R1kaB3rN

Nixpkgs PR https://github.com/NixOS/nixpkgs/pull/381975

### DIFF
--- a/packaging/nix/old-unwrapped.nix
+++ b/packaging/nix/old-unwrapped.nix
@@ -1,0 +1,60 @@
+{
+  # Dependencies
+  lib,
+  cargo,
+  python3Packages,
+  rustPlatform,
+  umu-launcher-unwrapped,
+  # Public API
+  withTruststore ? true,
+  withDeltaUpdates ? true,
+}:
+umu-launcher-unwrapped.overridePythonAttrs (prev: {
+  # The nixpkgs patches (in `prev.patches`) are not needed anymore
+  # - no-umu-version-json.patch was resolved in:
+  #   https://github.com/Open-Wine-Components/umu-launcher/pull/289
+  # - The other is backporting:
+  #   https://github.com/Open-Wine-Components/umu-launcher/pull/343
+  patches = [];
+
+  nativeCheckInputs =
+    (prev.nativeCheckInputs or [])
+    ++ [
+      python3Packages.pytestCheckHook
+    ];
+
+  nativeBuildInputs =
+    (prev.nativeBuildInputs or [])
+    ++ [
+      rustPlatform.cargoSetupHook
+      cargo
+    ];
+
+  pythonPath = with python3Packages;
+    (prev.pythonPath or [])
+    ++ [
+      urllib3
+      (callPackage ./pyzstd.nix {})
+    ]
+    ++ lib.optionals withTruststore [
+      truststore
+    ]
+    ++ lib.optionals withDeltaUpdates [
+      cbor2
+      xxhash
+    ];
+
+  configureFlags =
+    (prev.configureFlags or [])
+    ++ [
+      "--use-system-pyzstd"
+      "--use-system-urllib"
+    ];
+
+  preCheck = ''
+    ${prev.preCheck or ""}
+
+    # Some tests require a writable HOME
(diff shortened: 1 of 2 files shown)


R1kaB3rN:
Great, the relevant pull request was finally merged in nixpkgs. @LovingMelody are you fine/have any strong feelings for these changes? 

If not, then I can go ahead and merge this.

packaging/nix/old-unwrapped.nix
```diff
@@ -0,0 +1,60 @@
...
+    ++ [
+      rustPlatform.cargoSetupHook
+      cargo
+    ];
+
+  pythonPath = with python3Packages;
+    (prev.pythonPath or [])
+    ++ [
+      urllib3
+      (callPackage ./pyzstd.nix {})
```
R1kaB3rN, line 37: I believe we can drop this once it lands in unstable.

packaging/nix/unwrapped.nix
```diff
@@ -1,79 +1,57 @@
...
+    # Specify ourselves which tests are disabled
+    disabledTests = [
+      # Broken? Asserts that $STEAM_RUNTIME_LIBRARY_PATH is non-empty
+      # Fails with AssertionError: '' is not true : Expected two elements in STEAM_RUNTIME_LIBRARY_PATHS
+      "test_game_drive_empty"
+      "test_game_drive_libpath_empty"
+
+      # Broken? Tests parse_args with no options (./umu_run.py)
+      # Fails with AssertionError: SystemExit not raised
+      "test_parse_args_noopts"
```
R1kaB3rN, line 55: I understand why the game drive tests fail but am not sure for this one. I suppose it's related to the other tests failing, and I am not too interested in looking into it.

packaging/nix/old-unwrapped.nix
MattSturgeon: If we're confident that we have an up-to-date nixpkgs revision, then the entire `old-unwrapped.nix` file can go, along with the pyzstd package.

It is only kept to allow for a smoother transition for end users who may end up bumping their umu input but their nixpkgs input remains slightly behind.

We should be able to drop it in a few weeks. Maybe throwing an error if a v1.1.4 base package is still found.

R1kaB3rN:
Aside: If either pyzstd or zstd get updated, which will happen soon, then I imagine it will require a rebuild for the upstream nixpkg, correct?

MattSturgeon:
> Aside: If either pyzstd or zstd get updated, which will happen soon, then I imagine it will require a rebuild for the upstream nixpkg, correct? 

Typically nixpkgs will update shortly after a new version is released. If umu still needs the old version, then nixpkgs may consider packaging both versions. Or the dependency can be overridden to use the old version, specifically for umu.

Otherwise, if umu requires the new version before it is available in nixpkgs, then we can apply overrides here.

How much effort we put in here would depend on how important it is for umu to have a specific version of zstd.

LovingMelody:
No strong feelings on this pr

R1kaB3rN (approved):

PR #384, opened 2025-02-15
build(nix): minor fixes

MattSturgeon:
Based on @R1kaB3rN's feedback in https://github.com/NixOS/nixpkgs/pull/381975

Pushed up separately to #374, so that we can keep that one small & atomic.

- Use configure flags to disable building vendored dependencies
- Ensure pystd is always installed - it isn't optional
- Switch from `propagatedBuildInputs` to `pythonPath`
- Move cargo lock to before dependencies
- Added check phase

### DIFF
--- a/packaging/nix/unwrapped.nix
+++ b/packaging/nix/unwrapped.nix
@@ -14,36 +14,16 @@ umu-launcher-unwrapped.overridePythonAttrs (prev: {
   src = ../../.;
   inherit version;
 
+  cargoDeps = rustPlatform.importCargoLock {
+    lockFile = ../../Cargo.lock;
+  };
+
   # The nixpkgs patches (in `prev.patches`) are not needed anymore
   # - no-umu-version-json.patch was resolved in:
   #   https://github.com/Open-Wine-Components/umu-launcher/pull/289
   # - The other is backporting:
   #   https://github.com/Open-Wine-Components/umu-launcher/pull/343
   patches = [];
-
-  # The `umu-vendored` target needs submodules. However, we don't actually need
-  # this target or those submodules anyway, since we add `pyzstd` as a nix package
-  #
-  # As a temporary solution, we explicitly specify the supported build targets:
-  buildFlags =
-    (prev.buildFlags or [])
-    ++ [
-      "umu-dist"
-      "umu-launcher"
-    ];
-
-  # Same issue for install targets
-  installTargets =
-    (prev.installTargets or [])
-    ++ [
-      "umu-dist"
-      "umu-docs"
-      "umu-launcher"
-      "umu-delta"
-      "umu-install"
-      "umu-launcher-install"
-      "umu-delta-install"
-    ];
 
   nativeBuildInputs =
     (prev.nativeBuildInputs or [])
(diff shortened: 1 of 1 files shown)


packaging/nix/unwrapped.nix
```diff
@@ -52,21 +38,42 @@ umu-launcher-unwrapped.overridePythonAttrs (prev: {
...
+
+    # Broken? Tests parse_args with no options (./umu_run.py)
+    # Fails with AssertionError: SystemExit not raised
+    "test_parse_args_noopts"
+  ];
+
+  preCheck = ''
+    ${prev.preCheck or ""}
+
+    # Some tests require a writable HOME
```
R1kaB3rN, line 76: I'll have to rethink and improve the tests later because we shouldn't have to write to $HOME. It may require making the relevant functions more testable by accepting $XDG_DATA_HOME as a new parameter instead of calling Path.home(), which should be easy to do.

packaging/nix/unwrapped.nix
```diff
@@ -52,21 +38,42 @@ umu-launcher-unwrapped.overridePythonAttrs (prev: {
...
-  };
+  configureFlags =
+    (prev.configureFlags or [])
+    ++ [
+      "--use-system-pyzstd"
+      "--use-system-urllib"
+    ];
+
+  disabledTests = [
+    # Broken? Asserts that $STEAM_RUNTIME_LIBRARY_PATH is non-empty
```
R1kaB3rN, line 63: If it's empty then it implies ldconfig does not exist or it does, but could not find the shared library paths of the system. This varies by distribution, and in Arch Linux it's /usr/lib and /usr/lib/libfakeroot at the very least. In addition to those, umu-launcher includes the game directory, but the expectation is for there to be at least 2 paths.

R1kaB3rN: For NixOS I can imagine it can vary across systems. For home manager, which I use, the path is:
```
STEAM_RUNTIME_LIBRARY_PATH=/usr/lib64:/usr/lib32:/nix/store/m2mwcap0d7ylsap005vpr6xan063jlax-glibc-2.40-66/lib
```

This was run with the command `PROTONPATH=GE-Proton GAMEID=umu-0 umu-run ""`

R1kaB3rN: I think the test is OK to be skipped here, and I'm sorta thinking an e2e test may be more appropriate instead of a unit test here...

packaging/nix/unwrapped.nix
```diff
@@ -52,21 +38,42 @@ umu-launcher-unwrapped.overridePythonAttrs (prev: {
...
+    # Fails with AssertionError: '' is not true : Expected two elements in STEAM_RUNTIME_LIBRARY_PATHS
+    "test_game_drive_empty"
+    "test_game_drive_libpath_empty"
```
R1kaB3rN, line 66: These tests failing are related to the above.

packaging/nix/unwrapped.nix
R1kaB3rN: In addition to not writing to $HOME, the unit tests needs to be refactored to creating temporary directories in the system's tmpfs whenever possible.

Both of those changes will be large and done at a later time in a separate pull request. Until that's merged, I'm not really comfortable with suggesting to run the unit tests in nixpkgs.

packaging/nix/unwrapped.nix
MattSturgeon: I suspect the issue here is that we run the tests within the same sandbox environment that's used to build the package.

Therefore impure environment variables, including things related to Steam, are not set.

This is an intentional design decision intended to keep nix packages "pure" and deterministic.

Is the test assuming it is running on a fully installed system?

packaging/nix/unwrapped.nix
MattSturgeon: I agree, unit tests writing to (or reading from) HOME is not ideal. It's inherently impure and will result in tests that are not pure or idempotent.

However since nix runs the tests in a sandbox and we have to use a temp HOME if we want it to be writable _anyway_, this isn't a blocker from nixpkgs perspective.

MattSturgeon: `writableTmpDirAsHomeHook` was [merged into nixpkgs](https://github.com/NixOS/nixpkgs/pull/370869) a couple weeks ago, so I'll use that in the downstream package. Could use it here too, but it requires bumping the lockfile.

R1kaB3rN: Sure. Let’s use writableTmpDirAsHomeHook then merge this PR afterward

R1kaB3rN (approved):

PR #401, opened 2025-02-25
umu_run: do not force the runtime for winetricks if `UMU_NO_RUNTIME` is set

loathingKernel:
Since #293 is still under discussion, I am re-opening this as an interim solution to using `winetricks` with `proton-cachyos`

As it is now, trying to run `winetricks` enforces the use of the runtime even if `UMU_NO_RUNTIME` is set. This causes `proton-cachyos` to not run in its expected environment.

PR #293 is meant to address this in a seamless manner without exposing development features, but until then `UMU_NO_RUNTIME` is the feature itself.


loathingKernel:
Closing in favor of #410

PR #345, opened 2025-01-20
Refactor nix flake, based on overriding nixpkgs

MattSturgeon:
Now that there's a package being maintained in nixpkgs, it can be used as a starting point for the nix package here.

This is technically a breaking change, because the packages made available by the overlay are different and because the internals of the packages are different. But I've made efforts to avoid breaking changes to the public-facing API where practical.

### Testing

If you have a nix flake setup as described in the [README](https://github.com/Open-Wine-Components/umu-launcher?tab=readme-ov-file#nixos), you can simply replace your `inputs.umu.url` url string with:
```
umu.url = "github:MattSturgeon/umu-launcher?dir=packaging/nix"
```

Also, you may want to remove any `follows` targeting the umu input, or alternatively run `nix flake update`. Either way, the nixpkgs input must have https://github.com/NixOS/nixpkgs/pull/369259 and https://github.com/NixOS/nixpkgs/pull/375588.

You may occasionally need to update your lockfile (`nix flake update`) to get the latest changes from this PR.

### Packages

- `umu-launcher` is now equivalent to the old `umu` and `umu-run` packages (`combine.nix` & `umu-run.nix`)
- `umu-launcher-unwrapped` is now equivalent to the old `umu-launcher` (`umu-launcher.nix`)
- `umu-launcher-unwrapped` is directly based on nixpkgs, overriding the relevant attrs to use this git repo instead of 1.1.4
- `umu-launcher` uses the overridden `umu-launcher-unwrapped` implicitly

### Public API

The existing package-args API is still available, however warnings will be printed if it is used.

It consists of `override`-able package args:
- `version`
- `truststore`
- `cbor2` (although this never had any effect)

A new API is proposed:
- `version`
- `withTruststore`
- `withDeltaUpdates`

The README has been updated to reflect the new package-args API.

### Version

The package version is now based on the flake's `shortRev`, or `dirtyShortRev` if the flake is used locally and there are uncommitted changes.

This means users no longer need to override the version themselves.

The README has been updated to reflect this.

### Nixpkgs

The new implementation relies on nixpkgs having the `umu-launcher-unwrapped` package added in https://github.com/NixOS/nixpkgs/pull/369259 and a small fix from https://github.com/NixOS/nixpkgs/pull/375588.

That package has been in:
- master since 2024-01-19
- nixpkgs-unstable since 2024-01-20
- nixos-unstable-small since 2024-01-20
- nixos-unstable since 2024-01-22

The nixpkgs maintainers are:
- @diniamo
[...]

PR #402, opened 2025-02-25
build(nix): drop support for outdated nixpkgs revisions

MattSturgeon:
## Summary

#374 added support for nixpkgs revisions containing umu-launcher 1.2.0+, this PR drops support for nixpkgs revisions with older versions of the umu-launcher package.

## Changes

- drops `pyzstd.nix`
- drops `old-unwrapped.nix`
- adds `versionAtLeast "1.2.0"` assertion

## New assertion

The new assertion will likely remain indefinitely, with its minimum version being bumped whenever the downstream nixpkgs package adds some implementation detail that our flake package depends on. If the downstream package ever adds something we depend on _without_ bumping its version, the assertion would have to check that explicitly, e.g. by checking what attributes exist on the package.

As an example of the error message users would see, you can view this PR's [CI failure](https://github.com/Open-Wine-Components/umu-launcher/actions/runs/13533196761/job/37819769138?pr=402#step:5:219) from before the lockfile was updated :grin: 

## Grace period

This PR was initially a draft, because we want to grant end-users a grace period where their nixpkgs and umu inputs are allowed to be out-of-sync.

The grace period here shouldn't need to be very long though, unlike the renames done by #345. Those renames could trip up users who don't update their inputs often; some users likely still haven't updated their inputs since before #345.

In contrast, a user updating their umu input without updating their nixpkgs input is less likely. Also, the assertion message should be enough to detect and help resolve the issue.

---


MattSturgeon:
1.2.0+ has been available in all unstable nixpkgs channels since 2025-02-24, so just over a week now.

IDK how long we wanted the grace period to be, where users are allowed to have out-of-sync umu & nixpkgs inputs, but I'm going to undraft this now.

As outlined in the description above, the grace period for _this_ does not need to be anywhere near as long as the renames introduced in #345.

LovingMelody:
Did pyzstd get back ported to stable? I don't think umu is in stable either until 25.05 is released

MattSturgeon:
> Did pyzstd get back ported to stable? I don't think umu is in stable either until 25.05 is released

This PR isn't dropping _stable_ nixpkgs support, because that was never supported to begin with.

The flake requires an unstable nixpkgs input that contains an overridable `umu-launcher` package, regardless of this PR.

What the PR drops is support for nixpkgs unstable inputs that contained the _older_ umu-launcher package where more things needed to be overridden. The newer umu-launcher package in nixpkgs is basically identical to this flake, so we're only overriding `src`.

PR #410, opened 2025-03-05
umu_run: handle Protons without an explicit runtime requirement

loathingKernel:
Handle Proton versions that do not set a runtime in their `toolmanifest.vdf`. Such versions include non-runtime versions such as proton-cachyos and Proton from Valve before version 5.13.

For now they are executed on the host system outside of the runtime.

Supersedes #401

### DIFF
--- a/tests/test_config.sh
+++ b/tests/test_config.sh
@@ -19,5 +19,6 @@ store = 'gog'
 " >> "$tmp"
 
 
+# This works only for an existing prefix, the prefix `~/Games/umu/umu-0` is created in the previous workflow steps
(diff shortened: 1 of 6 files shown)


GloriousEggroll:
Noting that this was reverted because I did not realize it was allowing proton to be run without the runtime container.

As noted per discord conversation:

we want to use the container and always use a default runtime if one is not specified. the whole point is to have a distro agnostic target -- which is what the container and runtime provide

reverted the changes for now from that pr. i think the best approach is going to be to use the default current runtime if one isnt specified like we already do . we definitely want to use the container + runtime at all times

we don't want people reporting bugs when they arent using the container. we simply cannot afford the manpower or time to debug a moving (or several moving) targets. long term we want to build inside the container for other applications (such as strider was doing for retroarch)

PR #427, opened 2025-03-20
Reapply "umu_run: handle Protons without an explicit runtime requirement"

loathingKernel:
After further discussion on Discord wrt #410, I am re-opening this with an few changes.

Specifically, umu will allow a tool without a runtime requirement to run only if `UMU_NO_RUNTIME=1` is set, otherwise it will complain that it cannot resolve the required runtime. This ensures that the user is absolutely certain they want to run the tool without a runtime and they accept full responsibility if something doesn't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,13 @@ Borderlands 3 from EGS store.
 3. In our umu unified database, we create a 'title' column, 'store' column, 'codename' column, 'umu-ID' column. We add a line for Borderlands 3 and fill in the details for each column.
 4. Now the launcher can search 'Catnip' and 'egs' as the codename and store in the database and correlate it with Borderlands 3 and umu-12345. It can then feed umu-12345 to the `umu-run` script.
 
+## Reporting issues
+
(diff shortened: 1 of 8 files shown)


moutazhaq:
As someone who regularly uses custom Proton versions (like proton-cachyos) without the Steam runtime, I would really appreciate it if this functionality was added back to umu. This will allow me to keep using launchers like Lutris to run non-Steam games with those Proton versions.

Selmack:
I too would like for this functionality to return to the umu launcher.  I tend to use cachy's proton in conjunction with Heroic and Lutris launchers.

GloriousEggroll:
> This will allow me to keep using launchers like Lutris to run non-Steam games with those Proton versions.

Just pointing out you can already use Lutris to run non-steam games as-is even with the runtime. lutris uses umu to run any proton versions selected. the only difference there is whether the runtime provides wine its libraries or if the system does, and if the system is doing it (runtime disabled), we have 0 control over if something breaks or doesnt work correctly.

GloriousEggroll:
closing, no longer needed per internal conversation

PR #447, opened 2025-04-27
fix steam remote-play, remote-play together and gamerecording 

nodscher:
Steam expects the `SteamGameId` environment variable to stay unchanged in order for anything using Video-Capture to work, like remote-play, remote-play together and gamerecording. Thus we should leave it alone if it's set. 

In that case (or generally) SteamGameId is not supposed to equal SteamAppId.

this fix is neccessary for https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4493

### DIFF
--- a/umu/umu_run.py
+++ b/umu/umu_run.py
@@ -225,8 +225,7 @@ def set_env(
     if match(r"^umu-[\d\w]+$", env["UMU_ID"]):
         env["STEAM_COMPAT_APP_ID"] = env["UMU_ID"][env["UMU_ID"].find("-") + 1 :]
     env["SteamAppId"] = env["STEAM_COMPAT_APP_ID"]
-    env["SteamGameId"] = env["SteamAppId"]
-
+    env["SteamGameId"] = os.environ.get("SteamGameId", env["SteamAppId"])
     # PATHS
     env["WINEPREFIX"] = str(pfx)
     env["PROTONPATH"] = str(protonpath)
--- a/umu/umu_test.py
+++ b/umu/umu_test.py
@@ -2561,12 +2561,6 @@ class TestGameLauncher(unittest.TestCase):
                 self.env["STEAM_COMPAT_APP_ID"],
                 "Expected SteamAppId to be STEAM_COMPAT_APP_ID",
             )
-            self.assertEqual(
-                self.env["SteamGameId"],
-                self.env["SteamAppId"],
-                "Expected SteamGameId to be STEAM_COMPAT_APP_ID",
-            )
-
             # PATHS
             self.assertEqual(
                 self.env["STEAM_COMPAT_SHADER_PATH"],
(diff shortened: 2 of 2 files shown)


loathingKernel:
If I understand correctly, the goal is to keep whatever random number steam sets here for gamerecording etc to work. Problem is Proton's wine predominantly checks `SteamGameId` to apply its own fixes. If we allow steam's random number for non-steam games, wouldn't we lose fixes from wine's code potentially?

nodscher:
Oh. I didn't know that. 
The only why around that without modifying proton I could come up with is saving SteamGameId to a seperate Variable and monkey patching `subprocess.call` using `PYTHONPATH` to reapply it when Proton is done.
Would something like this be acceptable?
```
#sitecustomize.py

import subprocess
import os
from functools import wraps

original_call = subprocess.call

@wraps(original_call)
def patched_call(*args, **kwargs):
    try:
        kwargs["env"]["SteamGameId"] = os.environ["SavedSteamGameId"]
    except: 
        pass

    return original_call(*args, **kwargs)

subprocess.call = patched_call
```
But there might be some fatal flaw with this that I am missing and there are probably other ways?

nodscher:
I also just noticed that the AppId needed for the gamescope_baselayer stuff is exactly `SteamGameId/4294967296` across all non-steam apps I've tried on both my Steamdeck an PC. So as long as that number is not specific to my account saving the SteamGameId will allow umu to work with shader-caching disabled within gamescope (I think).

Etaash-mathamsetty:
> If I understand correctly, the goal is to keep whatever random number steam sets here for gamerecording etc to work. Problem is Proton's wine predominantly checks `SteamGameId` to apply its own fixes. If we allow steam's random number for non-steam games, wouldn't we lose fixes from wine's code potentially?

Steam should automatically set both SteamAppId and SteamGameId equal to each other for steam games (I believe, ill have to check)

loathingKernel:
> Steam should automatically set both SteamAppId and SteamGameId equal to each other for steam games (I believe, ill have to check(

It does, but most commonly it checks the "SteamGameId" env variable. There are ~59 checks using `SteamGameId` and ~7 using `SteamAppId`, and that's why I brought it up.

There is always the option of using `sed` to quickly replace them at build time with `SteamAppId`, or maintaining a patch to rename them, neither of which is preferable imo. Both have their own issues, either upkeeping the patch with new changes or sed looking a bit like magic for anyone wanting to implement umu support for their proton.

Hence personally I consider both of these as a last resort.

nodscher:
I don't think maintaining that monkeypath (or do you mean a .patch file?) would be a problem. From looking at Protons Github Repo the same call has been used since proton started using that python file, so I think it's unlikely to change (and even if, it shouldn't break anything). 
I also liked that approach because it doesn't change anything in the Proton files. But it seems kinda sketchy so I'm unsure if it's fine.

loathingKernel:
> I don't think maintaining that monkeypath (or do you mean a .patch file?) would be a problem.

If this refers to me, I was referencing a potential patch file to replace all `SteamGameId` with `SteamAppId` in wine's code.

Etaash-mathamsetty:
I feel like something minor about the SteamAppId/SteamGameId within Proton should not block this PR, and that we should try to get it merged. There are many ways to work around the SteamAppId/SteamGameId issues by modifying wine. Also, there are a tiny number of games that would need this.

PR #450, opened 2025-05-04
umu_run: create `pfx/creation_sync_guard` file to satisfy proton requirements for partial prefixes

loathingKernel:
umu-launcher creates a partial prefix to control some aspects of its folder structure. With Proton 10, since when creating a new prefix the `version` file doesn't exist yet, proton assumes there is no need to upgrade the prefix and `upgrade_pfx()` doesn't run. This has the side-effect of the `pfx/creation_sync_guard` file not being created and proton assumes the prefix was not created properly. As a result it tries to back it up and continue, which also negates any kind of control umu can have on the prefix.

See: https://github.com/ValveSoftware/Proton/blob/proton_10.0/proton#L923-L933

### DIFF
--- a/umu/umu_run.py
+++ b/umu/umu_run.py
@@ -77,6 +77,7 @@ def setup_pfx(path: str) -> None:
         pfx.symlink_to(Path(path).expanduser().resolve(strict=True))
 
     Path(path).joinpath("tracked_files").expanduser().touch()
+    Path(path).joinpath("creation_sync_guard").expanduser().touch()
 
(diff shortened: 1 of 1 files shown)


loathingKernel:
Due to https://github.com/ValveSoftware/Proton/commit/89c962511833b6148fd1e4e38d35a405465e94fa, this is not relevant any more. Leaving it open as a reminder that this needs to be resolved somehow. Unsure of how presently.

PR #476, opened 2025-05-24
allow STEAM_COMPAT_MOUNTS to be set by the user

nodscher:


R1kaB3rN:
Unnecessary. A method to share more paths to the container already exists.

See 
- https://github.com/Open-Wine-Components/umu-launcher/wiki/Frequently-asked-questions-(FAQ)#how-do-i-make-more-files-available-to-the-steam-runtime-container 
-  https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/docs/shared-paths.md#usually-not-shared

PR #479, opened 2025-05-29
feat: add support for launching foreign arch executables inside a container runtime

R1kaB3rN:


GloriousEggroll:
we need a check somewhere to make sure its not being run by fex or box64 -- specifically fedora 42 auto-tries to run x86 executables using fex via binfmt, and i know playtron uses box64

R1kaB3rN:
I think we should wait for Valve first before merging this because it looks like they will be taking a different approach for running x86 exes on aarch64 systems.

Specifically, it appears there will be a different compatibility tool with the [app ID 3810310](https://steamdb.info/app/3810310/) to support that use case. Also, Steam will be expected to configure pressure-vessel via `STEAM_COMPAT_EMULATOR` and `STEAM_COMPAT_GRAPHICS_PROVIDER` when necessary.

Issue #318, opened 2025-01-02
UMU Heroic Steam Deck - Every Game


RawToast225:
Every single game that I try to play on Heroic in gaming mode with any version of proton (I've mostly only tried the last two) doesn't work, but those same games work correctly in desktop mode.

When I load a game in game mode, it stays on the steam game loading screen and plays the sound of the game, but does not move on to the game screen. It eventually gives me the option to press B to abort the game, so I tend to do that after a bit.

This is the Heroic log for Cat Quest that does not work properly in game mode:
[051eaac0842c46d7a5a62858ad534d5a-lastPlay.log](https://github.com/user-attachments/files/18287425/051eaac0842c46d7a5a62858ad534d5a-lastPlay.log)

This is the Heroic log for Cat Quest that does work in desktop mode with no other changes:
[051eaac0842c46d7a5a62858ad534d5a-lastPlay(1).log](https://github.com/user-attachments/files/18287428/051eaac0842c46d7a5a62858ad534d5a-lastPlay.1.log)

My hope is that I've configured something incorrectly on my system, but for the life of me I can't imagine what it would be. I use Bazzite on my steam deck, as the log shows, and it still updates properly, so it's passing their chucksums.

To be clear, this happens with every single game that I try to play through Proton-GE+UMU in game mode, but it works with only Proton-GE in game mode.

R1kaB3rN:
The symptoms tells me that, for some reason, none of the requirements were met for umu-lancher to fix windows. I don't know the details on how Bazzite is configured, but, for the most part, as long as there's an X server and that distro isn't mucking with gamescope environment variables or DISPLAY outputs, then it should work in HGL. At least on SteamOS.

R1kaB3rN:
Also, in this issue tracker, please use the umu-launcher tool directly if you're looking to get any support. You can do this by simply starting a shell inside the Flatpak sandbox then by running the executable. Just be sure to pass Heroic's environment variables as well. For example:

```
flatpak run --command=sh --devel --filesystem=$(pwd) $FLATPAK_ID
```
```
[...]

PR #491, opened 2025-06-16
derive steam_appid from SteamGameId

nodscher:
From my testing, the appId that the steammode code expects is always exactly `(SteamGameId - 2²⁵) / 2³²`.

So, I think that after bit shifting by 32bits, it should be usable as a fallback when shader pre-caching is disabled.

### DIFF
--- a/umu/umu_run.py
+++ b/umu/umu_run.py
@@ -504,6 +504,9 @@ def get_steam_appid(env: MutableMapping) -> int:
         with suppress(ValueError, IndexError):
             return int(Path(path).parts[-2])
 
+    with suppress(ValueError):
+            return int(env.get("SteamGameId", "")) >> 32
+
     return steam_appid
 
 


R1kaB3rN:
Thanks, this is great :smile:. 

Confirmed that this is the case on my end too. We can also use `SteamOverlayGameId`, which is observed to have the same value, but I don't think it's necessary as `SteamGameId` is more well understood since it's used throughout Valve's wine source code.

cc @RawToast225 and @termdisc from https://github.com/Open-Wine-Components/umu-launcher/issues/318.

umu/umu_run.py
```diff
@@ -504,6 +504,9 @@ def get_steam_appid(env: MutableMapping) -> int:
         with suppress(ValueError, IndexError):
             return int(Path(path).parts[-2])
 
+    if gameId := env.get("SteamGameId"):
+            return int(gameId)>>32
```
R1kaB3rN: It's safe to use the [] operator here because umu-launcher is guaranteed to set SteamGameId.

```
    if gameId := env["SteamGameId"]:
            return int(gameId)>>32
```

R1kaB3rN: Since commit 8eb75f449871d91c97f25cfec7d611b5510c0598, umu-launcher will prioritize the one from the environment. It may/may not be set by the Steam client. Its value may/may not be an integer either (e.g., starcitizen). So I think this should be handled. For example:

```
    with suppress(ValueError):
        return int(env["SteamGameId"]) >> 32

```

nodscher: CI complains without get (https://github.com/Open-Wine-Components/umu-launcher/actions/runs/15705917308/job/44251562874?pr=491)

R1kaB3rN (approved):

R1kaB3rN:
The wiki has been updated. Until the next release, Steam Deck users who prefer to have Steam's Shader Pre-Caching disabled can get the new build at https://github.com/Open-Wine-Components/umu-launcher/actions/runs/15719366905/artifacts/3349719838.

Etaash-mathamsetty:
how did you figure this out btw

nodscher:
> how did you figure this out btw

I assumed that there must me some connection between the IDs, so I put some in a spreadsheet and their quotient was always almost 2³². And after comparing the numbers in binary I saw that the first 32bits were always the AppId.

Issue #470, opened 2025-05-20
On the Steam Deck, with umu enabled in heroic games launcher flatpak, steam deck UI (controller configuration, power menu, etc) disappears and the game is "always on top"


pizzadude:
With umu enabled in heroic games launcher flatpak, steam deck UI (controller configuration, power menu, etc) disappears and the game is "always on top".

It seems with umu enabled, on the steam deck with the heroic games launcher flatpak, the steam deck UI gets hidden often. I can't configure controls while the game is open because the game keeps pushing itself to the top. Happens in any game with umu enabled.

pizzadude:
Here's a video of the issue. https://www.youtube.com/watch?v=ouwR1Y0OxyE

The problem happens around 1:50.

Notice the Steam Deck's UI disappears when I start to select controller settings, and I can't bring it back up (the steam menu or the other menu). Also, I can only get back to the Steam Deck's UI briefly,by bringing up the virtual keyboard, but if I go back to controller settings or any other menu it disappears again.

I'm on SteamOS Preview if that matters. (but I don't think it does, as this doesn't happen with UMU disabled)

arielj:
one comment, in the meantime (until the fix is released), you can bring back the steam menu and the power settings menu if you press `Steam+X` to open the keyboard, that will move the layers and will put those 2 overlays back over the game as a workaround.

the fix linked there fixes the problem

R1kaB3rN:
A fix for this issue was merged. Until the next release, users can get it by downloading the artifact at: https://github.com/Open-Wine-Components/umu-launcher/actions/runs/15269383018/artifacts/3201668301.

pizzadude:
Thank you.

PR #496, opened 2025-06-24
only run in steammode when running in a flatpak container

nodscher:
the steammode code only needs to run when in a flatpak. Running that anyway can cause Problems if there are bugs in UMU (for example #470 also was an Issue when using Heroics AppImage, even though the Steam Reaper could have handled that properly on it's own)

### DIFF
--- a/umu/umu_run.py
+++ b/umu/umu_run.py
@@ -632,10 +632,11 @@ def run_command(command: tuple[Path | str, ...]) -> int:
         os.environ.get("XDG_CURRENT_DESKTOP") == "gamescope"
         or os.environ.get("XDG_SESSION_DESKTOP") == "gamescope"
     )
+    is_flatpak: bool = (os.environ.get("container") == "flatpak") # noqa: SIM112
     # Note: STEAM_MULTIPLE_XWAYLANDS is steam mode specific and is
     # documented to be a legacy env var.
     is_steammode: bool = (
-        is_gamescope_session and os.environ.get("STEAM_MULTIPLE_XWAYLANDS") == "1"
+        is_gamescope_session and os.environ.get("STEAM_MULTIPLE_XWAYLANDS") == "1" and is_flatpak
     )
 
     if not command:


umu/umu_run.py
```diff
@@ -632,10 +632,11 @@ def run_command(command: tuple[Path | str, ...]) -> int:
         os.environ.get("XDG_CURRENT_DESKTOP") == "gamescope"
         or os.environ.get("XDG_SESSION_DESKTOP") == "gamescope"
     )
+    is_flatpak: bool = (os.environ.get("FLATPAK_ID") is not None)
```
R1kaB3rN: Would prefer:
```
    is_flatpak: bool = os.environ.get("container") == "flatpak"
```

Flatpak follows the Container Interface outlined by systemd and we use that logic in other places too. See https://systemd.io/CONTAINER_INTERFACE/

R1kaB3rN (changes requested):

umu/umu_run.py
R1kaB3rN: Also, I don't think the issue is necessarily Flatpak-specific. I'll have to test a bit later, but I hypothesize that whenever umu is running in a sandbox that's configured a certain way, the issue of game windows not being focused will occur.

R1kaB3rN (approved):

umu/umu_run.py
nodscher: hm ok that would make sense. maybe: `is_flatpak: bool = os.environ.get("container") is not none`?

but I also just noticed that you can just launch another steam reaper inside the steamrt, so maybe we could just scrap all that code and let steam handle it?

R1kaB3rN: > hm ok that would make sense. maybe: is_flatpak: bool = os.environ.get("container") is not none?

It's too general and this issue has only been observed for Flatpak. Can't dismiss cases for containers that are properly configured. The logic wouldn't account for cases like by simply using `bwrap` to create a similarly constructed sandbox instead of using a framework like Flatpak.

> but I also just noticed that you can just launch another steam reaper inside the steamrt, so maybe we could just scrap all that code and let steam handle it?

Sure, would be interested in a specific example.

loathingKernel: > but I also just noticed that you can just launch another steam reaper inside the steamrt, so maybe we could just scrap all that code and let steam handle it?

Sorry to intervene, but what does reaper have to do with this issue? How does it come into the equation here?

nodscher: adding this after the rt:
```
"$STEAM_BASE_FOLDER/ubuntu12_32/reaper"\
    SteamLaunch \
    AppId=$appId --
```
    
I got this simplified bash wrapper that works
```
#!/bin/bash
echo "game: $1 with args: ${@:2}"

runtimedir=".local/share/Steam/"


appId=$(python -c "print(${SteamGameId}>>32)")
echo "Using AppId: $appId from SteamGameId: ${SteamGameId}"
echo "${DXVK_STATE_CACHE_PATH}"

export STEAM_COMPAT_CLIENT_INSTALL_PATH=$HOME/.local/share/Steam
export STEAM_COMPAT_APP_ID=0

if [[ -n $WINEPREFIX ]]; then
    export STEAM_COMPAT_DATA_PATH=${WINEPREFIX}
else
    export STEAM_COMPAT_DATA_PATH=$HOME/.local/share/Steam/steamapps/compatdata/"$appId"
fi
[ ! -d "${STEAM_COMPAT_DATA_PATH}" ] && mkdir -p "${STEAM_COMPAT_DATA_PATH}"
echo ${STEAM_COMPAT_DATA_PATH}

[[ -z "${PROTONPATH}" ]] && export PROTONPATH=$HOME/.local/share/Steam/compatibilitytools.d/GE-Proton10-3
echo "Using Proton path: ${PROTONPATH}"

export STEAM_COMPAT_MOUNTS=$HOME/$runtimedir/steamapps/common/SteamLinuxRuntime_sniper
export STEAM_COMPAT_PROTON=1
export STEAM_COMPAT_TOOL_PATHS="${PROTONPATH}":"$HOME"/$runtimedir/steamapps/common/SteamLinuxRuntime_sniper
export PROTON_CRASH_REPORT_DIR=/tmp/proton_crashreports
export STEAM_COMPAT_LIBRARY_PATHS=$HOME/$runtimedir/steamapps

"$HOME/.local/share/umu/steamrt3/umu"\
    --verb=waitforexitandrun -- \
    "$HOME/.local/share/Steam/ubuntu12_32/reaper"\
    SteamLaunch \
    AppId=$appId -- \
    "${PROTONPATH}/proton" \
    waitforexitandrun \
    "$1" "${@:2}"
```

nodscher: > > but I also just noticed that you can just launch another steam reaper inside the steamrt, so maybe we could just scrap all that code and let steam handle it?
> 
> Sorry to intervene, but what does reaper have to do with this issue? How does it come into the equation here?

as far as I can tell It is what steam uses for showing the apps

loathingKernel: Not really, `reaper` is simply there to have a process stay alive while the processes that spawn as children of it might exit, the reaper process will stay alive for as long as there are children. This is [`reaper`](https://github.com/Plagman/reaper/blob/main/reaper.cpp), it forks a child process that runs the target application and the parent process is just waiting for all the children ( and children of the children ) to exit. If that is the issue with umu, this PR https://github.com/Open-Wine-Components/umu-launcher/pull/434 is replicating `reaper` functionality but in umu, as normally umu will exit after the parent inferior process exits, while its children might still be active.

From what I understand about the steammode code, as I have never tested it myself, it is running a daemon thread, which monitors windows. That thread will exit whenever umu reaches the end of execution, which is tied to the parent inferior process as it is right now.

The runtime seems to offer reaper functionality in some cases through `steam-runtime-launcher-service` (I think this is the way it does it, but I am not sure) but this seems to not always be the case or it could be working differently on the Deck or in Flatpak. If for showing the application in gamescope (which is what prompted the PR I linked in the first place) a process is required to stay alive, then probably `umu` should be the one that is kept alive.

nodscher: Ok that makes sense, but why does Steam give the reaper those Arguments and why does executing a new reaper with them bring the game to the front?

loathingKernel: If you are referring to this 
```bash
"$HOME/.local/share/umu/steamrt3/umu"\
    --verb=waitforexitandrun -- \
    "$HOME/.local/share/Steam/ubuntu12_32/reaper"\
    SteamLaunch \
    AppId=$appId -- \
    "${PROTONPATH}/proton" \
    waitforexitandrun \
    "$1" "${@:2}"
```
then it is in the order of the arguments, you execute `reaper` in the Steam Runtime, before `proton`. I should mention here that the `proton` script in the case of `umu` calls `wine start.exe <application>` which doesn't wait for the application to complete unless you run it with the `/wait` argument (we tried that at some point but it caused other issues), and the proton script proceeds to exit. By inserting the `reaper` binary there, the calling process from the runtime is kept alive and as a result umu also stays alive, basically having the same effect as if the reaper was before the call to `"$HOME/.local/share/umu/steamrt3/umu"`. This would be actually interesting to test if it works in this case as it would be the equivalent to
```
/home/loathingkernel/.local/share/Steam/ubuntu12_32/reaper SteamLaunch AppId=204100 --
/home/loathingkernel/.local/share/Steam/ubuntu12_32/steam-launch-wrapper --
'/home/loathingkernel/.local/share/Steam/steamapps/common/SteamLinuxRuntime_sniper'/_v2-entry-point
```

The arguments are likely there to be read later on through the `proc` interface or some other facility that can look at the application arguments. They might or might not be required. and they also might have a role in this whole thing.

nodscher: I had tested all that (if I understand you right) and from my testing it only brings it to the front with the reaper after the runtime and with those arguments set
(though when escaping the flatpak via flatpak-spawn it doesn't seem to matter where the reaper is in the chain, but it definitely needs a new one)

nodscher: I don't know why it works but I have made this branch with all the manual x/gamexcope code removed and the steam reaper added and it does work: https://github.com/nodscher/umu-launcher/tree/steamReaper
(build at https://github.com/nodscher/umu-launcher/actions/runs/15865586679)

loathingKernel: I see what you mean more accurately after reviewing the branch you linked. This is really interesting. Using `reaper` from Steam would add an implicit dependency on Steam, but arguably we could ship it with umu again as he have done in the past. Could you test with the following subreaper implementation, in terms of functionality it should be 1:1 with the repo I linked above. It could also help reveal if Steam's current `reaper` does something more that we are not aware of. It should be a drop in replacement.

```python
#!/usr/bin/env python3

import logging
import os
import sys
from argparse import Namespace
from ctypes import CDLL, c_int, c_ulong
from ctypes.util import find_library
from logging import getLogger
from typing import List

# Constant defined in prctl.h
# See prctl(2) for more details
PR_SET_CHILD_SUBREAPER = 36


def get_libc() -> str:
    """Find libc.so from the user's system."""
    return find_library("c") or ""


def subreaper(args: Namespace, other: List[str]) -> int:
    logger = getLogger("subreaper")
    logging.basicConfig(
        format="[%(name)s] %(levelname)s: %(message)s",
        level=logging.DEBUG if args.debug else logging.INFO,
        stream=sys.stderr,
    )

    logger.debug("command: %s", args)
    logger.debug("arguments: %s", other)

    command: List[str] = [args.command, *other]
    workdir: str = args.workdir
    child_status: int = 0

    libc: str = get_libc()
    prctl = CDLL(libc).prctl
    prctl.restype = c_int
    prctl.argtypes = [
        c_int,
        c_ulong,
        c_ulong,
        c_ulong,
        c_ulong,
    ]
    prctl_ret = prctl(PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0, 0)
    logger.debug("prctl exited with status: %s", prctl_ret)

    pid = os.fork()  # pylint: disable=E1101
    if pid == -1:
        logger.error("Fork failed")

    if pid == 0:
        sys.stdout.flush()
        sys.stderr.flush()
        os.chdir(workdir)
        os.execvp(command[0], command)  # noqa: S606

    while True:
        try:
            child_pid, child_status = os.wait()  # pylint: disable=E1101
            logger.info("Child %s exited with wait status: %s", child_pid, child_status)
        except ChildProcessError as e:
            logger.info(e)
            break

    return child_status


if __name__ == "__main__":
    sep = sys.argv.index("--")
    argv = sys.argv[sep+1:]
    args = Namespace(command=argv.pop(0), workdir=os.getcwd(), debug=True)
    subreaper(args, argv)


__all__ = ["subreaper"]
```

This should be equivalent to compiling and using `reaper` from [the repo](https://github.com/Plagman/reaper) but easier to include and iterate upon.

nodscher: Your python reaper doesn't work but the C one does, so Steams Reaper doesn't seem to do anything special. Passing on those Arguments in the way the C reaper does seems important though? But I really don't know. 

And wouldn't a dependency on Steam be fine if it's explicitly only used when umu is run from Steam?

loathingKernel: > Your python reaper doesn't work but the C one does

I know that is does what it is supposed to do in terms of reaping the children, so something else might be going on here, maybe the executable name is important or something similar, going off by this description in the original repo. How did you use it?

`Anything before -- is ignored on the command-line, which can be used to annotate a given process tree with metadata for external parties to inspect. Thus, any process in the sub-command can reliably be associated with such metadata:`

> Passing on those Arguments in the way the C reaper does seems important though?

I think so yes, same reason as above.

> And wouldn't a dependency on Steam be fine if it's explicitly only used when umu is run from Steam?

Could be, but it could also be useful elsewhere. It would also solve the issue I described in https://github.com/Open-Wine-Components/umu-launcher/pull/434

nodscher: But I think I found why the Problem happens in the first place. Reaper seems to lose everything in the runtime and is thus not associating it with the metadata required for steam. I have no clue what metadata even means here, but it's what the repo says.

flatpak pstree:
```
reaper─┬─bwrap───bwrap─┬─bwrap───heroic
       │               ├─2*[cat]
       │               └─heroic-run───heroic─┬─heroic───heroic───34*[{heroic}]
       │                                     ├─heroic───9*[{heroic}]
       │                                     ├─heroic───20*[{heroic}]
       │                                     ├─python3───steam-runtime-l───2*[{steam-runtime-l}]
       │                                     ├─zypak-sandbox
       │                                     └─46*[{heroic}]
       └─bwrap───xdg-dbus-proxy───3*[{xdg-dbus-proxy}]
```
with flatpak-spawn:
```
reaper─┬─bwrap───bwrap─┬─bwrap───heroic
       │               ├─2*[cat]
       │               └─heroic-run───heroic─┬─flatpak_escape───flatpak-spawn───3*[{flatpak-spawn}]
       │                                     ├─heroic───heroic───34*[{heroic}]
       │                                     ├─heroic───9*[{heroic}]
       │                                     ├─heroic───20*[{heroic}]
       │                                     ├─zypak-sandbox
       │                                     └─46*[{heroic}]
       └─bwrap───xdg-dbus-proxy───3*[{xdg-dbus-proxy}]
```
native:
```
reaper───heroic─┬─heroic───heroic───34*[{heroic}]
                ├─heroic───heroic
                ├─heroic───12*[{heroic}]
                ├─heroic───20*[{heroic}]
                ├─umu-run───srt-bwrap───pv-adverb─┬─plugplay.exe
                │                                 ├─services.exe───4*[{services.exe}]
                │                                 ├─svchost.exe───2*[{svchost.exe}]
                │                                 ├─umu-shim───python3───wine64
                │                                 ├─wineboot.exe
                │                                 ├─winedevice.exe───5*[{winedevice.exe}]
                │                                 └─wineserver
                └─45*[{heroic}]
```

nodscher: > How did you use it?

I used the umu branch I linked and set the steam_base_folder so that it uses the c reaper I compiled/ your python script exactly as the one provided by steam

loathingKernel: I think metadata refers to this information 
```
    SteamLaunch \
    AppId=$appId -- \
```
and it appears to be important to gamescope https://github.com/ValveSoftware/gamescope/blob/master/src/steamcompmgr.cpp#L4578-L4610
So the process name has to match `reaper` and contain the two metadata values `SteamLaunch` and `AppId=xxxx`

Edit: coincidentally this is also why the python reaper didn't work, its process name is `python3`
call:
```
UMU_LOG=1 PROTONPATH=umu-sniper umu-run ../../raredevs/Rare/rare/commands/subreaper.py asda asdasd adsasd -- sleep 5000
```
pstree:
```
systemd(1,1)───alacritty(43279,43265)───bash(43286,43286)───umu-run(44435,44435)───srt-bwrap(44439,44439)───pv-adverb(44521,44521,mnt,user)───python3(44548,44521)───sleep(44550,44521)
````

nodscher: yeah I got that it means those, I meant that I don't know what (or rather where?) that metadata is. I only found in the proc/cmdline of reaper.

~~and I did call the python wrapper reaper, so that should be matched?~~
you're right it's called python3

loathingKernel: I've updated the script to rename the process, I guess the only unknown that remains to be seen is if the metadata is discovered nicely by `gamescope`

https://gist.github.com/loathingKernel/bc37a17380cbf29ce295f5b8855f0831

nodscher: it works
gamescope brings it to the front

loathingKernel: That's great. I was looking for a way to integrate this into umu itself without an external program, but it doesn't seem possible. While we could convince `umu-run` to pretend it is `reaper` in a similar way as above, I can't think of a way  to "pretend" our `/proc/PID/cmdline` contains something that `gamescope` will parse. I guess this was not for nothing though as it serves as documentation of how it works.

The issue that remains to be figured out is why it doesn't work with flatpak if it is prepended and not set after the runtime, if I am not mistaken. In that regard, I have no issue with adding either the C reaper or the python reaper in the command chain, since it also would simplify umu's code. I do have a preference towards the python one though since it doesn't require us to compile it and it is easier to package.

For posterity, previously I mentioned that Steam Runtime's pressure vessel has its own subreaper, it's part of the facilities of `pv-adverb` and enabled when using the `--subreaper` arguments, which seems to be the default behavior in any case.

nodscher: I'm not that familiar with C, but from what I can tell looking at the source code, you can't give pressure vessels subreaper any arguments.

Pressure vessel does have `steam-runtime-launch-options`, which seems to add a new reaper with the arguments, but I think it's is never run by pressure vessel

I assume the reason for it not working when it's prepended is the same as why steam's own reaper doesn't work. But I don't know why that is. I think it probably works when reaper is in the runtime because it's in the same namespace (or whatever else is different) as the Game.

My guess for it breaking in the first place would be that everything in the flatpak doesn't have access to the new processes (or pid space?), but it works from the host because that can still access pressure vessels child processes. Especially because the same happens with flatpak-spawn. But I could be completely wrong.

PR #497, opened 2025-06-25
use an external reaper and let gamescope handle running in steammode

nodscher:
In order to bring a window to the front gamescope expects the processes pid 1 to be a process called `reaper`, which has `SteamLaunch AppId=appid` in its cmdline.

Normally Steams own reaper handles that, but when running a steamrt from within a flatpak that reaper can't see everything in the bwrap, causing them to lose that metadata. This causes https://github.com/ValveSoftware/gamescope/issues/1341.

This adds a python subreaper implementation written by @loathingKernel (they also generally helped a lot with this) which re-adds the metadata to all child processes, causing them to appear again, without any of manual code umu currently runs.

This also means that we could completely reliably get the AppId from the cmdline of umus pid 1 (when run from steam), but I have no clue how to do that.

Theoretically we could also use steams built in reaper or this [reaper](https://github.com/Plagman/reaper) (probably the same as steams).

A thread about this change and stuff is in #496

### DIFF
--- a/umu/umu_consts.py
+++ b/umu/umu_consts.py
@@ -29,13 +29,6 @@ class HTTPMethod(Enum):
     TRACE = "TRACE"
 
 
-class GamescopeAtom(Enum):
-    """Represent Gamescope-specific X11 atom names."""
-
-    SteamGame = "STEAM_GAME"
-    BaselayerAppId = "GAMESCOPECTRL_BASELAYER_APPID"
-
-
 class FileLock(Enum):
     """Files placed with an exclusive lock via flock(2)."""
 
--- a/umu/umu_reaper.py
+++ b/umu/umu_reaper.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+
+import logging
+import os
+import sys
+from argparse import Namespace
+from ctypes import CDLL, byref, c_int, create_string_buffer
+from ctypes.util import find_library
+from logging import getLogger
+
+# Constant defined in prctl.h
+# See prctl(2) for more details
+PR_SET_CHILD_SUBREAPER = 36
+PR_SET_NAME = 15
+
+def get_libc() -> str:
+    """Find libc.so from the user's system."""
+    return find_library("c") or ""
+
+
+def subreaper(args: Namespace, other: list[str]) -> int:  # noqa: D103
+    logger = getLogger("subreaper")
+    logging.basicConfig(
+        format="[%(name)s] %(levelname)s: %(message)s",
+        level=logging.DEBUG if args.debug else logging.INFO,
+        stream=sys.stderr,
+    )
+
+    logger.debug("command: %s", args)
+    logger.debug("arguments: %s", other)
+
+    command: list[str] = [args.command, *other]
+    workdir: str = args.workdir
+    child_status: int = 0
+
+    libc: str = get_libc()
+    prctl = CDLL(libc).prctl
+    prctl.restype = c_int
+    prctl.argtypes = [
+        c_int,
+        # c_ulong,
+        # c_ulong,
+        # c_ulong,
+        # c_ulong,
+    ]
+
+    proc_name = b"reaper"
+    buff = create_string_buffer(len(proc_name)+1)
+    buff.value = proc_name
+    prctl_ret = prctl(PR_SET_NAME, byref(buff), 0, 0, 0)
+    logger.debug("prctl PR_SET_NAME exited with status: %s", prctl_ret)
+
+    prctl_ret = prctl(PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0, 0)
+    logger.debug("prctl PR_SET_CHILD_SUBREAPER exited with status: %s", prctl_ret)
+
+    pid = os.fork()  # pylint: disable=E1101
+    if pid == -1:
+        logger.error("Fork failed")
+
+    if pid == 0:
+        sys.stdout.flush()
+        sys.stderr.flush()
+        os.chdir(workdir)
+        os.execvp(command[0], command)  # noqa: S606
+
+    while True:
+        try:
+            child_pid, child_status = os.wait()  # pylint: disable=E1101
+            logger.info("Child %s exited with wait status: %s", child_pid, child_status)
+        except ChildProcessError as e:
+            logger.info(e)
+            break
+
+    return child_status
+
+
+if __name__ == "__main__":
+    sep = sys.argv.index("--")
+    argv = sys.argv[sep+1:]
+    debug = os.environ.get("UMU_REAPER_DEBUG") in {"1", "debug"}
+    args = Namespace(command=argv.pop(0), workdir=os.getcwd(), debug=debug)  # noqa: PTH109
+    subreaper(args, argv)
+
+
+__all__ = ["subreaper"]
--- a/umu/umu_run.py
+++ b/umu/umu_run.py
@@ -1,11 +1,9 @@
 import os
 import re
 import sys
-import threading
 from _ctypes import CFuncPtr
 from argparse import Namespace
-from array import array
-from collections.abc import Generator, MutableMapping
+from collections.abc import MutableMapping
 from concurrent.futures import Future, ThreadPoolExecutor
 from contextlib import suppress
 from ctypes import CDLL, c_int, c_ulong
@@ -23,10 +21,6 @@ from urllib3 import PoolManager, Retry
 from urllib3.exceptions import MaxRetryError, NewConnectionError
 from urllib3.exceptions import TimeoutError as TimeoutErrorUrllib3
 from urllib3.util import Timeout
-from Xlib import X, Xatom, display
-from Xlib.error import DisplayConnectionError
-from Xlib.protocol.request import GetProperty
-from Xlib.xobject.drawable import Window
 
 from umu import __runtime_versions__, __version__
 from umu.umu_consts import (
@@ -35,7 +29,6 @@ from umu.umu_consts import (
     STEAM_COMPAT,
     UMU_LOCAL,
     FileLock,
-    GamescopeAtom,
 )
 from umu.umu_log import log
 from umu.umu_plugins import set_env_toml
@@ -47,7 +40,6 @@ from umu.umu_util import (
     has_umu_setup,
     is_installed_verb,
     unix_flock,
-    xdisplay,
 )
 
 NET_TIMEOUT = 5.0
@@ -295,6 +287,8 @@ def build_command(
     shim: Path = local.joinpath("umu-shim")
     proton: Path = Path(env["PROTONPATH"], "proton")
     entry_point: Path = local.joinpath("umu")
+    reaper: Path = local.joinpath("reaper")
+    appId: int = get_steam_appid(env)
 
     if env.get("UMU_NO_PROTON") != "1" and not proton.is_file():
         err: str = "The following file was not found in PROTONPATH: proton"
@@ -319,6 +313,10 @@ def build_command(
             "--verb",
             env["PROTON_VERB"],
             "--",
+            reaper,
+            "SteamLaunch",
+            "AppId=" + str(appId),
+            "--",
             proton,
             env["PROTON_VERB"],
             env["EXE"],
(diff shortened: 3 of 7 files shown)


loathingKernel:
Do you think it would be possible or beneficial to keep the reaper script as a source file with umu's source and copy it over to the expected location? Furthermore, it might be helpful to have a rudimentary version check for the script to get it updated if required. For example if the file exists in the target location, you could checksum both files and if they don't match, copy over the one from the current version of umu or create the file again.

nodscher:
> Do you think it would be possible or beneficial to keep the reaper script as a source file with umu and copy it over to the expected location? Furthermore, it might be helpful to have a rudimentary version check for the script to get it updated if required. For example if the file exists in the target location, you could checksum both files and if they don't match, copy over the one from the current version of umu or create the file again.

I actually wanted to do that but couldn't think of an elegant way to do that.

But I'm probably just missing something. How should I do that?

loathingKernel:
I was thinking of something simple, like keeping the script as `umu_reaper.py` under `umu/`. Then in umu's source you can use the path of the current file to calculate the path to the `umu_reaper.py`
```python
file = Path(__file__).parent.resolve().joinpath("umu_reaper.py")
shutil.copyfile(file, local / "reaper")
```
if you don't want to copy the file (due to packaging or permissions issues maybe), you could read the file and write it
```python
file = Path(__file__).parent.resolve().joinpath("umu_reaper.py")
contents = file.read_text()
```
and then write the file as you do already. I didn't test the code so might contain some omission, but should reflect the idea.

I am not sure though if this could cause any packaging issues, but they could be worked around by treating the file as a resource rather than source code by the build system. Hopefully @R1kaB3rN can chime in here on their preferred way as they are more knowledgeable of umu's packaging challenges.

nodscher:
that doesn't work in the zipapp as far as I can tell
python gives `NotADirectoryError: [Errno 20] Not a directory` when trying anythin with opening/reading/copying files in there

loathingKernel:
I see, I have few more ideas based on python internals, but it'll diverge from the issue. Let's focus on the purpose of this PR for now and we can figure out my nitpick at a later stage.

loathingKernel:
I've made a small PoC of how this could work for the zipapp too <https://github.com/loathingKernel/reaper_factory>

Other than that, I think this is an important piece of the puzzle https://github.com/ValveSoftware/gamescope/blob/0d3a19791de447a673e315a5aed3f59ad64ebb71/src/steamcompmgr.cpp#L4638-L4670 and why we need this behaviour. so I wanted to include it in this PR for easy lookup.

loathingKernel:
Do we still need to override the `SteamGameId` with `SteamAppId` as done with https://github.com/Open-Wine-Components/umu-launcher/pull/447 or with this handling we can potentially pass `SteamAppId` to reaper and leave `SteamGameId` untouched? I guess there is *some* correlation here with reaper and gamescope using `AppId=` for the metadata.

Asking because some fixes are also setting the `SteamGameId` https://github.com/Open-Wine-Components/umu-protonfixes/blob/master/gamefixes-steam/1174180.py although I am not sure if this is intended or a forgotten artifact.

nodscher:
No that's separate, we still need the behavior from #447 for the Overlay. `SteamGameId` is needed for (and used by) the overlay and not gamescope (and the overlay doesn't seem to care about reaper metadata as far as I can tell).

And valve-wine and proton both only ever check either SteamGameId or SteamAppId, so I don't think it makes change sense to change the fixes.

loathingKernel:
Ah, I see, thanks for the clarification.

> And valve-wine and proton both only ever check either SteamGameId or SteamAppId, so I don't think it makes change sense to change the fixes.

I am not suggesting to change the fixes. What could be an issue here is that, due to the fixes hardcoding the `SteamGameId`, the Wine process will only know of that `SteamGameId`. What I care about in this instance is to know if this protonfixes behavior breaks something in what these PRs are trying to fix.

On the other hand, if this protonfixes behavior is not breaking anything, then it could be a solution to providing the proper `SteamGameId` to the game, despite what umu-launcher is exporting in its environment.

GloriousEggroll:
I've read through the comments here, and per the issue reported here (https://github.com/Open-Wine-Components/umu-launcher/issues/545#issuecomment-3639661817) if this resolves the problem with no issues I'm for merging it.

nodscher:
It unfortunately doesn't fix #545, as this PR is flatpak specific and the issue is specifacally about not being in flatpak.

But I have been using this approach (through a bash Script though) for bringing games to the front for the past 9 months on my SteamDeck and haven't had any issues that could happen here (only #545 through different means).

The only options I found for fixing #545 are:
- not overriding the SteamGameId (but this breaks wine hacks) (bad idea)
- reverting non-flatpak back to the manual layer code (reverting ea5a14c, which broke after 3c326d7) (this should definetely be done if nothing else is done)
- using that LD_PRELOAD I mentioned in https://github.com/Open-Wine-Components/umu-launcher/issues/545#issuecomment-3288532771 to give wine and gameoverlayrenderer a different SteamGameId (this fixes all problems I had with working together with Steam, including Game Recording/ Streaming and I have been using it for the past months without any issues)

I think this is still PR a better approach than what is done currently, but should wait for one of those fixes, because the amount of code removed would be very different

I also wanted to say for the record that the reaper loses the game in the first place, because whenever [srt sees a /.flatpkak-info it creates a subsandbox via the flatpak portal](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/pressure-vessel/wrap-flatpak.c), meaning it tells flatpak over the dbus to create a new sandbox (just like flatpak-spawn)

Here is that LD_PRELOAD code if anyone with C knowledge wants to do something with it:
```
#define _GNU_SOURCE
#include <dlfcn.h>
#include <stdlib.h>
#include <string.h>
#include <execinfo.h>
#include <stdio.h>
#include <stdint.h>

typedef char *(*orig_getenv_f_type)(const char *name);

typedef struct {
    uintptr_t start;
    uintptr_t end;
} lib_range_t;

static lib_range_t overlay_ranges[16];
static int overlay_count = 0;
static int overlay_ranges_initialized = 0;

static void init_overlay_ranges() {
    FILE *f = fopen("/proc/self/maps", "r");
    if (!f) return;

    char line[512];
    while (fgets(line, sizeof(line), f)) {
        if (strstr(line, "gameoverlayrenderer")) {
            uintptr_t start, end;
            if (sscanf(line, "%lx-%lx", &start, &end) == 2) {
                overlay_ranges[overlay_count].start = start;
                overlay_ranges[overlay_count].end = end;
                overlay_count++;
                if (overlay_count >= 16) break;
            }
        }
    }
    fclose(f);
    overlay_ranges_initialized = 1;
}

static int caller_is_overlay() {
    if (!overlay_ranges_initialized) {
        init_overlay_ranges();
    }

    void *buffer[32];
    int nptrs = backtrace(buffer, 32);

    for (int i = 0; i < nptrs; i++) {
        uintptr_t addr = (uintptr_t)buffer[i];
        for (int j = 0; j < overlay_count; j++) {
            if (addr >= overlay_ranges[j].start && addr < overlay_ranges[j].end) {
                return 1;
            }
        }
    }
    return 0;
}

static char *cached_umu_steam_game_id = NULL;
static int umu_steam_game_id_initialized = 0;

char *getenv(const char *name) {
    static orig_getenv_f_type orig_getenv = NULL;
    static char *cached_umu_steam_game_id = NULL;
    
    if (!orig_getenv) {
        orig_getenv = (orig_getenv_f_type)dlsym(RTLD_NEXT, "getenv");
    }

    if (strcmp(name, "SteamGameId") == 0) {

        if (!umu_steam_game_id_initialized) {
            char *val = orig_getenv("UMU_STEAM_GAME_ID");
            if (val) {
                cached_umu_steam_game_id = strdup(val);
            }
            umu_steam_game_id_initialized = 1;
        }

        if (cached_umu_steam_game_id != NULL) {
            if (caller_is_overlay()) {
                return cached_umu_steam_game_id;
            }
        }
    }

    return orig_getenv(name);
}
```

loathingKernel:
It might be worth noting here that gamescope also uses the value passed through `reaper AppId=<>` for its own hacks

https://github.com/ValveSoftware/gamescope/blob/08a4595b2bda6901792e2225a6b2f0c39266ebfd/src/steamcompmgr.cpp#L3287-L3351

nodscher:
Oh ): that's annoying. At least it's only 5 Games and 2 of them have the same thing worked around in wine

GloriousEggroll:
RE:
```
The only options I found for fixing https://github.com/Open-Wine-Components/umu-launcher/issues/545 are:

(1) not overriding the SteamGameId (but this breaks wine hacks) (bad idea)

(2) reverting non-flatpak back to the manual layer code (reverting https://github.com/Open-Wine-Components/umu-launcher/commit/ea5a14c9201e8ea15dd8d28360d2f58d9cacb7b0, which broke after https://github.com/Open-Wine-Components/umu-launcher/commit/3c326d71ef2380ce690c091212fe6fa705f9ef14) (this should definetely be done if nothing else is done)

(3) using that LD_PRELOAD I mentioned in https://github.com/Open-Wine-Components/umu-launcher/issues/545#issuecomment-3288532771 to give wine and gameoverlayrenderer a different SteamGameId (this fixes all problems I had with working together with Steam, including Game Recording/ Streaming and I have been using it for the past months without any issues)
```

Personally I'm for (2) revert ea5a14c9201e8ea15dd8d28360d2f58d9cacb7b0, would that be efficient enough paired with this PR to resolve #545 ?

GloriousEggroll:
I mean (3) also seems like a good choice if it ultimately fixes -everything-, I'm down for either. Maybe (2) -and- (3)?

loathingKernel:
@nodscher if we go with (3), where should the LD_PRELOAD take place? should it be outside the container or can it be inside? It's a matter of distribution, if it has to be outside, it needs to be part of umu-launcher, otherwise it could be part of umu-protonfixes, which gives us a predictable environment to build it in.

nodscher:
> Personally I'm for (2) revert [ea5a14c](https://github.com/Open-Wine-Components/umu-launcher/commit/ea5a14c9201e8ea15dd8d28360d2f58d9cacb7b0), would that be efficient enough paired with this PR to resolve #545 ?

(2) would fix it alone. The STEAM_GAME atom also just ends up as w->appId eventually, but for some reason doesn't care about SteamGameId. (https://github.com/ValveSoftware/gamescope/blob/08a4595b2bda6901792e2225a6b2f0c39266ebfd/src/steamcompmgr.cpp#L4602-L4615,  https://github.com/ValveSoftware/gamescope/blob/08a4595b2bda6901792e2225a6b2f0c39266ebfd/src/steamcompmgr.cpp#L5815-L5832)

You can just revert it now ig and if we decide to go with 3, I could just reapply it.

> I mean (3) also seems like a good choice if it ultimately fixes -everything-, I'm down for either. Maybe (2) -and- (3)?

It fixed all problems I had in that regard. A combination could also work where the AppId is still set manually, but the preload is used to make all the recording stuff work, but I presonally don't see the advantage of that.

>  if we go with (3), where should the LD_PRELOAD take place?

As far as I know it doesn't matter as long as the final (wine) executable has it. If it's set in umu Steam will associate a few more pids with the gameid, but I'm not aware of Steam doing anything with hose. 


I personally prefer only 3, because both Solutions are a bit fragile (#545 for the reaper implementation and https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4718, #470 for (2)), but the reaper is simpler codewise and the  LD_PRELOAD fixes gamerecording, even outside of gamescope (which may get more important with theater view streaming on the frame).

GloriousEggroll:
@nodscher alright lets go with (3), but I would also like to know a specific place to add it in umu

nodscher:
Sounds good.

Would you prefer the preload in umu-launcher or umu-protonfixes?
Would it being in umu-protonfixes mean that upstream Proton will still be broken?

If it should go into umu-launcher I would just add the env variable through the env dict.
For the binary itself I'm not sure. To work in the zipapp it probably needs to be in .local/share/umu. Optimally it could be copied there from the python blob without Internet. But I'm currently not sure on how to do that exactly (I would have to look into that).


Though I also noticed that at some point since September when I last fully tested the screencap stuff, Valve has started using `SteamOverlayGameId` in more places. The `Adding process A for gameID B` is now using it. Weirdly RemotePlay outside and game recording inside of the gamescope session has also started to work with `SteamOverlayGameId`, while RemotePlay inside and game recording outside of the gamescope session is still using `SteamGameId` (unless I was testing wrong). So with Valve changing stuff to use  `SteamOverlayGameId` there would also be a hope for a Option 4 of explicitly asking Valve to use `SteamOverlayGameId` in more places. That way we would only need the reaper.

Also (this is a bit besides the point but I wanted to document it):
I figured out why the reaper is "loosing the children" in the first place. Pressure Vessel cannot unshare the user NS further when inside a flatpak. So when SteamRuntime detects that it's running inside a flatpak it asks the flatpak portal over the dbus to create a "subsandbox" (It's doing essentially the same as the flatpak-spawn cli). It's not really a subsandbox though, but rather a sibling sandbox created by the flatpak portal (probably similar to nested containers in docker). Because there is no process-tree connection between the sandboxes, the new one will need a new reaper.
(https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/pressure-vessel/wrap-flatpak.c I think this is the most interesting piece of code for this, but SRT has a lot of code)

loathingKernel:
> For the binary itself I'm not sure. To work in the zipapp it probably needs to be in .local/share/umu. Optimally it could be copied there from the python blob without Internet. But I'm currently not sure on how to do that exactly (I would have to look into that).

We can not worry about that, we could dump the binary into a base64 string, store it as a python constant and write it back as binary on umu setup. It looks odd, and maybe a bit scary, but it will work and remain compatible with zipapp. The more important factor imo is to find the best possible solution, we can make it work afterwards.

> I figured out why the reaper is "loosing the children" in the first place. 

Good to finally have an explanation to this mess. Good to know, thank you.

GloriousEggroll:
made some quick tests with LD_PRELOAD cleared here:
https://github.com/Open-Wine-Components/umu-launcher/pull/620

games seem to launch fine in session now both when added as a non steam game from lutris and when run as a subprocess within lutris while lutris is running in the session as well.

closing this for now. can always revisit it if we need to change our decision

nodscher:
I don't really agree with unseting the LD_PRELOAD. Even if it's not that important in the gamescope session, it is still used for their fps-limiting.

And while we discussed the LD_PRELOAD here, this PR didn't really have anything directly to do with that. (LD_PRELOAD is only relevant outside of flatpak and this PR only in flatpak). I think this would still make sense to merge, especially if the manual xlib gameID code isn't to be used for flatpak-less gamescope sessions (no #579). (It would mean practically the same codepath in terms of window focus for both flatpak and non-flatpak)

Issue #339, opened 2025-01-18
Add "Launch Options" (not `launch_args`) to the `config.toml`


Kagukara:
Being able to add all your configurations to 1 file would great and what I thought the `config.toml` would allow, but it doesn't so far. 

There was this comment for a suggestion to add `[env]` for your environment variables https://github.com/Open-Wine-Components/umu-launcher/issues/135#issuecomment-2206921313 (hasn't had any activity since Jul 4, 2024) and this PR for adding `gamescope` https://github.com/Open-Wine-Components/umu-launcher/pull/179 that was closed (Oct 29, 2024).

Why not just add what Steam allows users to do with "Launch Options"? This would enable the user to add anything including environment variables and launch options to this section.

Example:
```toml
[umu]
launch_options = ["MESA_SHADER_CACHE_DIR=~/location/to/game/cache/mesa_shader", "VKD3D_SHADER_CACHE_PATH=/location/to/game/cache/vkd3d_shader", "gamemoderun", "gamescope -w 3840 -h 1800 -f"]
```

R1kaB3rN:
Yes. the current state of the configuration file usage leave much to be desired and hasn't been actively maintained or enhanced lately. The primary reason for this is simply because it's not clear how useful it really is. Since the Python rewrite and the initial release, no client app that I'm aware of uses config files and I'm observing the majority of users who use umu-launcher directly either make their own wrapper scripts or prefer the environment variable usage instead. The latter is understandable, as the usage is familiar to those who had previously used wine directly from the command line. For these reasons, the config usage is considered to be taken out in the next major release. At that time, users will be encouraged to look to other solutions such as [umu-wrapper](https://github.com/korewaChino/umu-wrapper) or build their own.

But if you think that adding an `[env]` table and putting all configurations to 1 file would make it more useful or if there's already client that's using config files, then I'm all ears. _Maybe_ we can keep it then.

For the new config file usage, maybe we can have the table structure like this:
```
[umu.umu-0]
proton = '~/.local/share/umu/compatibilitytools/GE-Latest'
prefix = '~/Games/umu/umu-0'
launch_args = ['foo', 'bar']
# This will prepend gamescope to the command line
gamescope = True

[umu.umu-1]
proton = '~/.local/share/umu/compatibilitytools/GE-Latest'
prefix = '~/Games/umu/umu-0'
launch_args = ['foo', 'bar']

# environment variables that will be applied globally
[umu.env]
[...]

Issue #130, opened 2024-06-29
is it possible to use umu inside bwrap


bohanubis:
basically I use nixos
but thats not really relevant I have this error
![image](https://github.com/Open-Wine-Components/umu-launcher/assets/77834479/a32cc14e-4994-4bdd-95fb-ee587942a343)

and I have no idea how to fix it
basically if someone was able just to run umu in bwrap just send the command

R1kaB3rN:
Yes, it's possible to use `bwrap` with umu as long as the sandbox is setup properly and the IPC namespace is shared. And I'm guessing you ran into the `KeyError` because your custom script runs umu under `bwrap` and is creating new namespaces? I don't know what's in your script, but I think unsharing the user namespace and setting the `--uid` option may be related to the `KeyError` as I expect that error to not happen under normal circumstances.

R1kaB3rN:
Also, I had thoughts to implement creating custom bwrap containers in the next release to allow modifying the filesystem (e.g., not exposing the entire home directory). However, this use case would be very niche and it would be in the configuration file usage. For example:

> [umu]
> game_id = 'flowers'
> exe = '/home/foo/Games/umu/flowers/drive_c/GOG Games/Flowers - Le Volume Sur Printemps/FLOWERS.exe'
> prefix = '/home/foo/Games/umu/flowers'
> proton = '/home/foo/.local/share/Steam/compatibilitytools.d/GE-Proton9-7'
> sandbox = true # Unshares the home directory and only exposes what's necessary
> 
>
> [[bwrap.bind]]  # Mounts the winetricks cache to the sandbox
> src = "/home/foo/.cache/winetricks"
> dest = "/home/foo/.cache/winetricks"

R1kaB3rN:
By the way, since it seems you use bwrap, umu sources it's own copy of bwrap so you actually don't need to install the bwrap package. You can find it in `$HOME/.local/share/umu/pressure-vessel/libexec/steam-runtime-tools-0/srt-bwrap`

bohanubis:
> Yes, it's possible to use `bwrap` with umu as long as the sandbox is setup properly and the IPC namespace is shared. And I'm guessing you ran into the `KeyError` because your custom script runs umu under `bwrap` and is creating new namespaces? I don't know what's in your script, but I think unsharing the user namespace and setting the `--uid` option may be related to the `KeyError` as I expect that error to not happen under normal circumstances.

can you give me a script with this actually working
this is mine
```bash
[...]

PR #563, opened 2025-10-29
Expand environemnt vars in config file

Jacajack:
I think having env var expansion in umu config files would be a very useful feature. I implemented a proof-of-concept which allows to express the paths in terms of each other. For example, `exe` can be specified relatively to `$WINEPREFIX`. I also implemented `UMU_CONFIG_FILE_DIR` which points to the directory containing the TOML config file. This is very handy if you want to keep the config file inside the wineprefix  and be able to move it easily without having to update all the paths.

```toml
[umu]
prefix = "${UMU_CONFIG_FILE_DIR}"
proton = "$WINEPREFIX/GE-Proton10-24"
exe = "$WINEPREFIX/drive_c/something.exe"
```

If you consider this something that could be added, I can work on updating the relevant docs and tests.

### DIFF
--- a/umu/umu_plugins.py
+++ b/umu/umu_plugins.py
@@ -63,9 +63,9 @@ def set_env_toml(
     env["UMU_CONFIG_FILE_DIR"] = str(config_path.parent)
 
     # Perform expansion of env vars for paths
-    env["WINEPREFIX"] = Template(env["WINEPREFIX"]).substitute(env)
-    env["PROTONPATH"] = Template(env["PROTONPATH"]).substitute(env)
-    env["EXE"] = Template(env["EXE"]).substitute(env)
+    env["WINEPREFIX"] = Template(env["WINEPREFIX"]).safe_substitute(env)
+    env["PROTONPATH"] = Template(env["PROTONPATH"]).safe_substitute(env)
+    env["EXE"] = Template(env["EXE"]).safe_substitute(env)
 
     _check_env(env)
 
--- a/umu/umu_plugins.py
+++ b/umu/umu_plugins.py
@@ -1,7 +1,7 @@
 from argparse import Namespace
 from pathlib import Path
 from typing import Any
-
+from string import Template
 
 def set_env_toml(
     env: dict[str, str], args: Namespace
@@ -49,7 +49,7 @@ def set_env_toml(
     with config_path.open(mode="rb") as file:
         toml = tomllib.load(file)
 
-    _check_env_toml(toml)
+    _check_toml(toml)
 
     # Required environment variables
     env["WINEPREFIX"] = toml["umu"]["prefix"]
@@ -59,6 +59,16 @@ def set_env_toml(
     env["GAMEID"] = toml["umu"].get("game_id", "")
     env["STORE"] = toml["umu"].get("store", "")
 
+    # Add config file dir to env so it can be referenced
+    env["UMU_CONFIG_FILE_DIR"] = str(config_path.parent)
+
+    # Perform expansion of env vars for paths
+    env["WINEPREFIX"] = Template(env["WINEPREFIX"]).substitute(env)
+    env["PROTONPATH"] = Template(env["PROTONPATH"]).substitute(env)
+    env["EXE"] = Template(env["EXE"]).substitute(env)
+
+    _check_env(env)
+
     if isinstance(toml["umu"].get("launch_args"), list):
         opts = toml["umu"]["launch_args"]
     elif isinstance(toml["umu"].get("launch_args"), str):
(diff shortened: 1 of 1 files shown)


Jacajack:
Perhaps I started reinventing the wheel a bit here. Just realized I could simply wrap `umu-run` in a bash script like this:

```bash
#!/bin/bash

export WINEPREFIX="$(dirname "$0")"
export PROTONPATH="$WINEPREFIX/GE-Proton10-24"
export GAMEID="umu-489830"
export STORE="gog"
export EXE="$WINEPREFIX/drive_c/GOG Games/Skyrim Anniversary Edition/Data/SkyrimTogetherReborn/SkyrimTogether.exe"

umu-run "$EXE" "$@"
```

and it will serve the same purpose, but without touching the Python code. In either case, I'm curious what's your view on that.

R1kaB3rN:
Yeah, so my view here is to just create a shell script if you require environment variables and want to _dynamically_ set values. Otherwise, _declare_ the path. So for the TOML keys that require a string that represent a file system path, my suggestion would be to declare the absolute path or a relative/symbolic one by using `~`, `..`, or `.` in the string if you want shorter ones. The `pathlib` module's parser will resolve the meaning of those characters.

---
<details>
<summary>Additional context/thoughts on configuration files</summary>

A bit of background though, the configuration file serves an alternative interface for umu-launcher. [Originally, the interface was designed to be simply the script's command](https://github.com/Open-Wine-Components/umu-launcher/blob/4cffc2c3a1348069f8be4e8fa774cab24ee9e28c/tools/umu-run-posix#L108) that's prepended and appended with the required inputs. For example:

```
WINEPREFIX=~/.wine umu-run foo.exe
```

This interface and usage replicates wine's, which makes `umu-run` intuitive and comfortable to use/implement for those already familiar with [wine(1)](https://man.archlinux.org/man/extra/wine/wine.1.en). Additionally, it's stateless; you don't need to persist a file containing those inputs to disk[1]. The issue though, is that sometimes you need or want to apply/experiment with a different set of inputs for either the same or different game.

For instance, pressure-vessel, Proton, and it's subprojects (e.g., bubblewrap, dxvk, etc.) can actually be configured in a lot different ways[2], and switching between different commands by searching the shell history isn't really an effective workflow for testing.

It's also error-prone. For instance, some commands may look similar, yet slightly different in the shell history, some of those environment values may contain very long strings that can contain a variable number of spaces characters, quotations[3], etc,. and it's easy to type a command wrong over the network after SSH'ing to a different machine. Furthermore, a terminal command isn't exactly a standard format that's easily human-readable, exportable, or replicatable across machines (or across clients). The configuration file addresses those concerns, or just provides another option for those who have a _taste_ for it.

The reality though, it's observed that most use cases are simple, don't need it, or just don't have that _taste_, which is why development has been paused for that interface. Like, just setting the path to the prefix and the proton to use suffices. Most don't set mile-long configuration options to run their games either, and clients like Lutris are able to set the required inputs without error. Therefore, it's a bit questionable how useful the feature is right now, making it more of an option that shines in niche cases and can assist in development/testing.

Though, if you disagree or have different thoughts, please share; I'd be interested in feedback or a different perspective.

1. Though, statelessness could be achieved for configs by having `umu-run` accept a TOML string.
2. Configuration is basic right now, but there were ideas for [bwrap](https://github.com/Open-Wine-Components/umu-launcher/issues/130#issuecomment-2198320730) and [environment variables](https://github.com/Open-Wine-Components/umu-launcher/issues/339#issue-2796589773). Through this usage, there's also potential for [install scripts](https://github.com/Open-Wine-Components/umu-protonfixes/issues/404#issuecomment-3369177018). Pull requests for them are welcome.
3. Since the Python rewrite, there are no shell-escape bugs. Any shell interpreter-related bugs that occur at runtime will be related to pressure-vessel, the compatibility tool, or winetricks.
</details>

Jacajack:
Thanks for the detailed answer and the historical perspective. I'm inclined to agree that implementing this feature in `umu-run` would be just redundant. A bunch of bash files can provide the same functionality and not complicate maintenance and testing of the launcher.

It also seems better to have all variable evaluation related stuff done by bash. One problematic (although stupid) config example would be:

```toml
proton = "some_path"
prefix = "${PROTONPATH}/prefix"
exe = "$WINEPREFIX/drive_c/whatever.exe"
```

This requires a specific order of evaluation to function correctly. Unless the vars would be expanded recursively... which then becomes a mess and something you definitely don't want to do. 

I suppose my initial enthusiasm to add this feature stems from the long legacy of my WINE prefix setups. I started using PlayOnLinux probably like 10 years ago and later migrated to a [custom WINE prefix manager](https://github.com/Jacajack/winechad/). Both of them search of a predefined directory (`wine_dir`) for x86 and amd64 WINE versions. It's generally a different UI for the same setup.

I was really happy when I found umu, because the TOML config feature sounded exactly like what I had. My first instinct was to incorporate umu into this PoL-shaped framework I had ended up with. I needed env vars in TOML to make that happen. Now I've realized I don't need that. I never even needed my custom prefix manager in the first place. All this has just been a long de-evolution process of PoL into a bunch of bash files and me approaching a simple problem from a weird angle.

I think we can close this now that we agree that this wouldn't be of much use. Thank you for maintaining this tool.

PR #575, opened 2025-12-01
build(nix): fix versionCheckHook, update nixpkgs, standardise versions

MattSturgeon:
### Summary

Fixes a build failure with recent nixpkgs and makes the package versioning clearer.

cc @beh-10257 @LovingMelody

---

### What changed

1. **Configure flake lock commits**  
   `nix flake lock` / `nix flake update` now use conventional commits when run with `--commit-lock-file`.

2. **Update nixpkgs input**  
   Pulls in the latest nixpkgs, as it hasn't been bumped in a while.

3. **Use date-based versioning**  
   Nix builds now use `1.3.0-unstable-YYYY-MM-DD` instead of commit hashes.  
   This aligns with [nixpkgs' version style](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#versioning) and enables the next commit fixing `versionCheckHook`...  
   The first part is read from `umu/__init__.py`, which is the canonical project version.

4. **Fix versionCheckHook**  
   Patch ensures the hook passes even though the Python package uses a slightly different version format.

5. **PEP-440 runtime version**  
   Python package `__version__` is now `${version}.devYYYYMMDD`, making the `--version` output more meaningful.

### Testing

The nix package can't be built when tests fail, so I initially based this PR on the commit before #293, due to the test suite currently being broken on `main`.

Sadly, that lead to conflicts, so this is now rebased on `main` and cannot easiliy be tested. One way to test is adding `doCheck = false` to the package. Another way is to rebase on #574 which aims to fix the testsuite.

---

### For users

- The version you see in Python (`__version__`) or via `--version` is now PEP-440 compatible and includes a date, e.g. `1.3.0.dev20251130`.  
- The version you see when inspecting the nix package attributes now follows the Nixpkgs style, e.g. `1.3.0-unstable-2025-11-30`.


MattSturgeon:
> Another way [to test] is to rebase on #574 which aims to fix the testsuite.

I've done this locally. This PR builds fine when rebased on that PR. Frustratingly, I had the same conflicts rebasing onto that as I did rebasing onto latest `main`, because #574 reverts the changes that conflict 😁

GloriousEggroll:
if you could give this a rebase im willing to get it merged, ty!

PR #584, opened 2025-12-26
Allow the user/developer to define the location of the umu and Steam folders

Kyuyrii:
I noticed that the location used when UMU is being used by a Snap is quite bad, so I'd like to change it.

The problem with the current location is that since the $HOME folder is considered to be the $HOME/.local/share folder, this generates the creation of folders like applications and desktop-directories, polluting the user's $HOME with folders that may not even be useful, and perhaps confusing the user, since a Steam folder is created when umu-run downloads umu-proton.

I chose the AppsFiles folder because it's the same one I use in Zordeer and Meganimus, and I'm trying to make a Snap version of Zordeer.

### DIFF
--- a/umu/umu_consts.py
+++ b/umu/umu_consts.py
@@ -78,8 +78,6 @@ if os.environ.get("container") == "flatpak":  # noqa: SIM112
         if os.environ.get("HOST_XDG_DATA_HOME")
         else Path.home().joinpath(".local", "share")
     )
-elif os.environ.get("SNAP"):
-    XDG_DATA_HOME: Path = Path(os.environ["SNAP_REAL_HOME"]).joinpath("AppsFiles", "UMU-launcher")
 else:
     XDG_DATA_HOME: Path = (
         Path(os.environ["XDG_DATA_HOME"])
@@ -96,6 +94,13 @@ UMU_CACHE: Path = XDG_CACHE_HOME.joinpath("umu")
 UMU_COMPAT: Path = XDG_DATA_HOME.joinpath("umu", "compatibilitytools")
 
 STEAM_COMPAT: Path = XDG_DATA_HOME.joinpath("Steam", "compatibilitytools.d")
+
+# Variables to define folders used by UMU
+if "UMU_FOLDER" in os.environ:
+    UMU_LOCAL = Path(os.environ["UMU_FOLDER"]).joinpath("umu")
+    UMU_COMPAT = Path(os.environ["UMU_FOLDER"]).joinpath("umu", "compatibilitytools")
+if "STEAM_COMPAT_FOLDER" in os.environ:
+    STEAM_COMPAT = Path(os.environ["STEAM_COMPAT_FOLDER"]).joinpath("Steam", "compatibilitytools.d")
 
 # Constant defined in prctl.h
 # See prctl(2) for more details
--- a/umu/umu_consts.py
+++ b/umu/umu_consts.py
@@ -79,7 +79,7 @@ if os.environ.get("container") == "flatpak":  # noqa: SIM112
         else Path.home().joinpath(".local", "share")
     )
 elif os.environ.get("SNAP"):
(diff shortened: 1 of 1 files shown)


Kyuyrii:
Since Heroic also uses the UMU launcher, when umu is launched in the Snap version of Heroic, the umu folder is created in the $HOME folder.

Kyuyrii:
I specified an exact location because I'm assuming choosing hidden folders might be a problem.

But if it isn't, I suppose it makes sense to allow umu-run within a Snap to use the XDG_DATA_HOME environment variable instead of forcing a fixed location.

loathingKernel:
FWIW the location is deliberately shared to avoid duplication of steam runtimes and protons between launchers and packaging methods.

Kyuyrii:
> FWIW the location is deliberately shared to avoid duplication of steam runtimes and protons between launchers and packaging methods.

In this case, because the $HOME folder is being treated as XDG_DATA_HOME, the Steam folder created by UMU is not used by any version of Steam.

And even if you allow UMU Snap to define XDG_DATA_HOME, each Steam version has its own folder, as far as I remember.

This objective is only achieved by using UMU outside of a Snap app and using only the native version of Steam.

loathingKernel:
> In this case, because the $HOME folder is being treated as XDG_DATA_HOME, the Steam folder created by UMU is not used by any version of Steam.

Indeed, that's true, but that seems like a different issue than setting `XDG_DATA_HOME` to an arbitrary directory different than flatpak or native.

Kyuyrii:
> > In this case, because the $HOME folder is being treated as XDG_DATA_HOME, the Steam folder created by UMU is not used by any version of Steam.
> 
> Indeed, that's true, but that seems like a different issue than setting `XDG_DATA_HOME` to an arbitrary directory different than flatpak or native.

I chose to change the location to $HOME/AppsFiles/UMU-launcher to avoid creating the applications and desktop-directories folders in the $HOME folder, as well as to avoid confusion about the Steam folder.

If someone opens the AppsFiles folder and sees a folder called UMU-launcher, they will immediately understand that the Steam folder is something from the UMU launcher and not something that Steam created.

I was very confused when I saw the Steam folder in the $HOME folder; it took me a while to understand that it was the UMU launcher that was creating it.

I know there's an option to let the user/developer define where the XDG_DATA_HOME will be, but I imagine that wasn't done due to the desktop-launch of the gnome and kde-neon-6 extensions.

Currently I'm testing the UMU launcher in 3 Snaps: zordeer, heroic, and umu-run.

If XDG_DATA_HOME is not set by the UMU launcher, XDG_DATA_HOME will be "$SNAP_USER_DATA/.local/share", so SteamRT3 will be downloaded 3 times.

loathingKernel:
I am probably missing something here, because I do not have any experience with Snap, but what is stopping us from setting the path to the following?
```python
    XDG_DATA_HOME: Path = Path(os.environ["SNAP_REAL_HOME"]).joinpath(".local", "share")
```

Kyuyrii:
> I am probably missing something here, because I do not have any experience with Snap, but what is stopping us from setting the path to the following?
> 
> ```python
>     XDG_DATA_HOME: Path = Path(os.environ["SNAP_REAL_HOME"]).joinpath(".local", "share")
> ```

The default behavior of Snap apps is to not have access to hidden folders, but there are some exceptions. I suppose the decision not to use $SNAP_REAL_HOME/.local/share was for that reason.

However, I can't say whether they didn't use that location because it caused some problem with UMU within Snap, or if they didn't use that location to avoid problems with other Snap apps.

Kyuyrii:
I made a modification that I think makes more sense.

It's now possible to define the location of the umu and Steam folders.

If it's not defined by the user or developer, it will use the default location, within XDG_DATA_HOME.

This solves the problem and gives more freedom.

GloriousEggroll:
looks good to me, thanks for taking the time to work on this. We were definitely trying at some point but kind of got put off by the heavy permission restrictions by upstream required to be accepted. Ive gone ahead and merged. fingers crossed!

PR #589, opened 2026-01-09
Only consider umu setup if there are files in .local/share/umu/

CommandMC:
When umu starts downloading the runtime required for the specified tool, it will create a folder for it in UMU_LOCAL. If this download fails, the following code is supposed to catch the error, explain it, and stop the launch:

https://github.com/Open-Wine-Components/umu-launcher/blob/24cdc86a1565764655c9de12404b2c5d12dec6e7/umu/umu_run.py#L888-L894

However, `has_umu_setup` currently only checks whether there is any non-lock file/folder in UMU_LOCAL, meaning the empty runtime directory that was created gets found, and umu is considered "setup".

With this PR, we instead search for non-lock files (not folders) in UMU_LOCAL, which avoids the above issue

### DIFF
--- a/umu/umu_util.py
+++ b/umu/umu_util.py
@@ -368,7 +368,7 @@ def extract_tarfile(path: Path, dest: Path) -> Path | None:
 def has_umu_setup(path: Path = UMU_LOCAL) -> bool:
     """Check if umu has been setup in our runtime directory."""
     return path.exists() and any(
-        file for file in path.glob("*") if not file.name.endswith("lock")
+        file for file in path.glob("**/*") if file.is_file() and not file.name.endswith("lock")
     )
 
 


R1kaB3rN (approved):

R1kaB3rN:
For additional context, one case where an empty subdirectory in `UMU_LOCAL` can occur is when a new user is on a network that’s online but repo.steampowered.com is blacklisted.

umu/umu_util.py
```diff
@@ -368,7 +368,7 @@ def extract_tarfile(path: Path, dest: Path) -> Path | None:
 def has_umu_setup(path: Path = UMU_LOCAL) -> bool:
     """Check if umu has been setup in our runtime directory."""
     return path.exists() and any(
-        file for file in path.glob("*") if not file.name.endswith("lock")
+        file for file in path.glob("**/*") if file.is_file() and not file.name.endswith("lock")
```
R1kaB3rN, line 371: I had time to take a bit of a closer look. While this change suffices, here are some thoughts. Feel free to ignore them.

---

When using Protons built against Valve's container runtimes, we know that umu has been setup if the contents of the Steam Linux Runtime tarball has been installed in an expected subdirectory (e.g., steamrt3) within umu-launcher's the base installation path. 

Rather than updating this function's internal logic to recursively list all the contents from the base path, checking the top-level contents of the relevant subdirectory for that game launch would make more sense:

```
                if not has_umu_setup(UMU_LOCAL.joinpath(runtime_variant)):
                    err: str = (
                        "umu has not been setup for the user\n"
                        "An internet connection is required to setup umu"
                    )
                    raise RuntimeError(err)
```

As a result, there'd be less files to check and other irrelevant subdirectories (e.g., steamrt4) are not searched when `has_umu_setup` is called when an `HTTPError` is raised.

CommandMC: Yes, I was thinking something like this as well. The question is how far we want to go with this, when exactly we consider umu "setup". Playing devils advocate, I would argue that *anything* being in the runtime directory would not count either, we'd have to make sure the files we expect to be there are (not) there

I believe a logical next step would be to verify that at least the `toolmanifest.vdf` file & the entry point listed there are present, although I don't know whether I have enough of a grasp of umu's codebase to implement a check for that cleanly

loathingKernel: Depends on how fast the check should be. If we just want to check for certain files, I think we can check for the following and be "sure enough"
* `<dir>/pressure-vessel/bin/pv-verify`
* `<dir>/VERSIONS.txt`
* `<dir>/_v2-entry-point`
* `<dir>/toolmanifest.vdf`
* `<dir>/mtree.txt.gz`

If we want a thorough check, we can run `pv-verify` to check the whole runtime using [`umu.umu_runtime.check_runtime()`](https://github.com/Open-Wine-Components/umu-launcher/blob/main/umu/umu_runtime.py#L338-L380)

R1kaB3rN: > Yes, I was thinking something like this as well. The question is how far we want to go with this, when exactly we consider umu "setup". Playing devils advocate, I would argue that _anything_ being in the runtime directory would not count either, we'd have to make sure the files we expect to be there are (not) there

Right, for an umu setup, we can consider performing either a simple or a deep check of the tarball contents on network error, which can occur both on install and on update. There are fews ways to handle this, and I don't have strong feelings of each one. Given your perspective on the problem, I'll leave the choice and implementation up to you.

Implementation aside, what I will primarily care about is the setup verification's performance, and problematic logic that would lead to bugs (e.g., false negatives).

R1kaB3rN, line 371: Recursively listing the contents from umu-launcher's base installation path can lead to a false negative in a session for users where a network errors occurs on runtime install, and there are multiple subdirectories (e.g., steamrt4) installed within umu-launcher's runtime installation path.

For example, if a client passes a Proton that requires a container runtime that hasn't been installed yet by umu-launcher, but during the installation process a network error had suddenly occurred.

R1kaB3rN (changes requested):

umu/umu_util.py
GloriousEggroll: I added a resolution for this by tracking runtime installs with a marker:

If a network failure occurs while installing (say) steamrt4, but steamrt3 is already installed, you no longer rely on a recursive directory walk (which can fail mid-iteration). Instead, you:

- backfill markers for already-installed runtimes once per session

- check for the specific runtime’s marker (UMU_LOCAL / runtime_variant) before deciding “umu isn’t set up”.

https://github.com/Open-Wine-Components/umu-launcher/commit/f75a33bb2af35b98a7e47d43635b9d7406037f4c

Issue #415, opened 2025-03-14
umu-launcher does not use network proxy


recreators01:
see title.
![Image](https://github.com/user-attachments/assets/c38c25a4-0176-41c4-8343-b1a13b5b0fd1)

curl output.
```
MECHREV% curl --proxy "127.0.0.1:7890" -I "https://github.com/"
HTTP/1.1 200 Connection established

HTTP/2 200 
server: GitHub.com
date: Fri, 14 Mar 2025 02:42:31 GMT
content-type: text/html; charset=utf-8
vary: X-PJAX, X-PJAX-Container, Turbo-Visit, Turbo-Frame, Accept-Language, Accept-Encoding, Accept, X-Requested-With
```

sonic-warrior:
Any update on this?

picklejump:
Bro, I think you can try modifying the umu code, replace urllib3.PoolManager with ProxyManager in site-packages/umu/umu_run.py

kingsidelee:
the urllib3 umu-launcher uses to download files does not get configuration from environment variables, it is intended behavior. [urllib3/urllib3/issues/1785](https://github.com/urllib3/urllib3/issues/1785#issuecomment-574644811)

if you do not want to modify the code, try using some tun based solution with custom route/redirect.

here is my workaround using [sing-box](https://github.com/SagerNet/sing-box) and linux netns to achieve proper isolation:

```
ip netns add proxy

# add veth for host and proxy namespace
ip link add veth-proxy-host type veth peer name veth-proxy-ns

# assume bridge br0 exist
ip link set veth-proxy-host master br0
ip link set veth-proxy-host up
ip link set veth-proxy-ns netns proxy

ip netns exec proxy ip link set lo up
ip netns exec proxy ip link set veth-proxy-ns up

# assume static ip assign
ip netns exec proxy ip addr add 192.168.1.128/24 dev veth-proxy-ns
ip netns exec proxy ip route add default via 192.168.1.1

# sing-box tun solution
ip netns exec proxy sing-box run

# cleanup
ip netns delete proxy
```

and run umu-launcher with oneliner
```
<your_envvars> sudo -E ip netns exec proxy sudo -E -u <your_uid> -g <your_gid> umu-run <rest_args>
```

picklejump:
@kingsidelee 
I think what the urllib3 official documentation says is correct: environment variables are defined at the operating system application level, not at the network layer. You should use requests instead of urllib3.
[...]

PR #650, opened 2026-04-09
feat: add support for https_proxy environment variable (#415)

recreators01:
`https_proxy` should be enough. 
Regarding the changes to `umu_runtime.py`, the URL must include a scheme when using `ProxyManager`; otherwise, it will not work.

### DIFF
--- a/umu/umu_run.py
+++ b/umu/umu_run.py
@@ -23,7 +23,7 @@ from tempfile import gettempdir
 from types import FrameType
 from typing import Any
 
-from urllib3 import PoolManager, Retry
+from urllib3 import PoolManager, Retry, ProxyManager
 from urllib3.exceptions import HTTPError
 from urllib3.util import Timeout
 from Xlib import X, Xatom, display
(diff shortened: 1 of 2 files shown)


loathingKernel:
Please rebase your PR properly on top of current `main` branch to avoid merge artifacts. Also avoid unnecessary style changes in unrelated code.

recreators01:
> Please rebase your PR properly on top of current `main` branch to avoid merge artifacts. Also avoid unnecessary style changes in unrelated code.

ok

recreators01:
Close:https://github.com/Open-Wine-Components/umu-launcher/issues/415
This also resolves the issue that URLs without a scheme (i.e., 'https://') are deprecated and will raise an error in urllib3 v3.0.

umu/umu_run.py
```diff
@@ -936,11 +936,19 @@ def _check_offline_runtime(_rt: UmuRuntime) -> bool:
     else:
         timeouts = NET_TIMEOUT
 
+    http_pool: PoolManager
+    if "https_proxy" in os.environ:
+        http_pool = ProxyManager(
+            proxy_url=os.environ["https_proxy"],
```
loathingKernel, line 942: Is lowercase `https_proxy` a common env variable or something you came up with for this PR? If it isn't common, IMO it would better to be uppercase, and specific to UMU, i.e. `UMU_HTTP_PROXY` to match the existing `UMU_HTTP_RETRIES` and `UMU_HTTP_TIMEOUT`

recreators01: Common env variable. Most app use it, e.g. curl wget git python urllib/urllib2.

loathingKernel: I see, for future reference this is the rationale curl provides: https://everything.curl.dev/usingcurl/proxies/env.html#http_proxy-in-lower-case-only

loathingKernel (approved):

imLinguin (approved):

PR #696, opened 2026-07-16
Don't resolve executable symlinks

JuneStepp:
Right now, the logic for ensuring the executable path is absolute, also resolves any symlinks. That can cause the game to be launched from the wrong directory, potentially giving confusing errors.
My use case involves symlinks from the Nix store, but I'm sure there are many other reasons a Linux user might use symlinks.

### DIFF
--- a/umu/umu_run.py
+++ b/umu/umu_run.py
@@ -253,7 +253,8 @@ def set_env(
             # Ensure executable path is absolute, otherwise Proton will fail
             # when creating the subprocess.
             # e.g., Games/umu/umu-0 -> $HOME/Games/umu/umu-0
-            exe: Path = Path(cmd).expanduser().resolve(strict=True)
+            exe: Path = Path(cmd).expanduser().absolute()
+            _ = exe.resolve(strict=True)
             env["EXE"] = str(exe)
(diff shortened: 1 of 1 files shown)


loathingKernel:
I have a vague memory of a requirement to have the path to the executable as absolute for some reason. I'll try to remember more and update here.

Edit: Maybe I don't need to remember, the following is the commented right above your change
>            # Ensure executable path is absolute, otherwise Proton will fail
>            # when creating the subprocess.

Edit2: Ok, I see now, the problem is the symlink getting resolved, not being absolute, but retaining the following line
```
            _ = exe.resolve(strict=True)
```
kind of threw me off, is there any reason this is kept as it is?

JuneStepp:
> Ok, I see now, the problem is the symlink getting resolved, not being absolute, but retaining the following line
> 
>             _ = exe.resolve(strict=True)
> 
> kind of threw me off, is there any reason this is kept as it is?

From the Python docs for `Path.resolve()`:

> If a path doesn’t exist or a symlink loop is encountered, and strict is True, [OSError](https://docs.python.org/3/library/exceptions.html#OSError) is raised. If strict is False, the path is resolved as far as possible and any remainder is appended without checking whether it exists.

That line is so the error is still raised when relevant.

